PR #39, opened 2021-08-19
What if `lifted_getfield` on a tangent just called `getproperty`

oxinabox:
Semantically these are the same.
but lets see if this works out right

### DIFF
--- a/src/stage1/generated.jl
+++ b/src/stage1/generated.jl
@@ -389,10 +389,7 @@ end
 lifted_getfield(x::ZeroTangent, s) = ZeroTangent()
 lifted_getfield(x::NoTangent, s) = NoTangent()
 
-function lifted_getfield(x::Tangent, s)
-    z = getfield(ChainRulesCore.backing(ChainRulesCore.canonicalize(x)), s)
-    z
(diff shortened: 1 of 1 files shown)


simeonschaub:
@keno Any objections? This should be fine, right?

Keno:
If it works, ok by me.

simeonschaub:
I plan to set up some CI as soon as I figure out the problem on latest Julia nightly. Since this doesn't seem urgent, I'll probably hold off on this until then, so we can merge with some confidence.

simeonschaub:
Ah, this breaks down for `getfield(::NamedTuple, ::Int)`. Good that we have CI now.

oxinabox:
We should consider adding that overload in CRC so this works.
It would be useful in other places.
(I wanted it in a Zygote PR where for NT I couldn't tell if I was going to get a symbol or a int)

simeonschaub:
@Keno Any ideas about the hang on Windows? Is that worth investigating?

Keno:
No idea. I've never tried it on windows. I'd say just disable the CI job for now.

PR #45, opened 2021-09-09
Add `RuleConfig`

mcabbott:
This wants to add methods for `rrule_via_ad` and `frule_via_ad` from ChainRules. 

~~Need to figure out how to test this; so far `rrule_via_ad` does seem to work.~~ Not sure anything is in the right place.

### DIFF
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -1,4 +1,5 @@
 using ChainRulesCore
+struct DiffractorRuleConfig <: RuleConfig{Union{HasReverseMode,HasForwardsMode}} end
 
 @Base.constprop :aggressive accum(a, b) = a + b
 @Base.constprop :aggressive accum(a::Tuple, b::Tuple) = map(accum, a, b)
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -13,7 +13,7 @@ first_partial(x::CompositeBundle) = map(first_partial, getfield(x, :tup))
 
 # TODO: Which version do we want in ChainRules?
 function my_frule(args::ATB{1}...)
-    frule(map(first_partial, args), map(primal, args)...)
+    frule(DiffractorRuleConfig(), map(first_partial, args), map(primal, args)...)
 end
 
 # Fast path for some hot cases
@@ -117,6 +117,8 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
         return shuffle_base(r)
     end
 end
+
+ChainRulesCore.frule_via_ad(::DiffractorRuleConfig, args...) = ∂☆internal{1}()(args...)
 
 function (::∂☆shuffle{N})(args::AbstractTangentBundle{N}...) where {N}
     ∂☆p = ∂☆{minus1(N)}()
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -119,10 +119,9 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
 end
 
 function ChainRulesCore.frule_via_ad(::DiffractorRuleConfig, partials, args...)
-    tangents = map(partials, args) do p, a
-        TangentBundle{1}(a, (p,))
-    end
-    ∂☆internal{1}()(tangents...)
+    bundles = map((p,a) -> TangentBundle{1}(a, (p,)), partials, args)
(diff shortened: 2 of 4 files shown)


simeonschaub:
Testing should now be set up, so you'll probably want to rebase this.

simeonschaub:
Would it be possible to add a test specifically making use of this?

mcabbott:
Yes was just trying to figure out how (and where). I've added something, but my `frule_via_ad` definition is clearly completely wrong.

src/runtime.jl
```diff
@@ -1,5 +1,7 @@
 using ChainRulesCore
 
+struct DiffractorRuleConfig <: RuleConfig{Union{HasReverseMode,HasForwardsMode}} end
```
mcabbott, line 3: This doesn't seem a great fit for the terse style of this package, better ideas are welcome.

simeonschaub: You mean the name? Seems fine to me.

mcabbott: Yes, sorry somehow I lost the word.

test/runtests.jl
```diff
@@ -1,8 +1,8 @@
 using Diffractor
-using Diffractor: var"'", ∂⃖
+using Diffractor: var"'", ∂⃖, DiffractorRuleConfig
 using ChainRules
 using ChainRulesCore
-using ChainRules: ZeroTangent, NoTangent
+using ChainRules: ZeroTangent, NoTangent, frule_via_ad, rrule_via_ad
```
oxinabox, line 5: These should be loaded from ChainRulesCore

oxinabox (approved):
You can test this using ChainRulesTestUtils if you wanted.
It supports testing these cases.

This basically LGTM.
I see no reason not to merge it once tests pass

src/stage1/forward.jl
```diff
@@ -118,6 +118,13 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
     end
 end
 
+function ChainRulesCore.frule_via_ad(::DiffractorRuleConfig, partials, args...)
+    tangents = map(partials, args) do p, a
+        TangentBundle{1}(a, (p,))
+    end
+    ∂☆internal{1}()(tangents...)
+end
```
mcabbott, line 126: I don't see what's going wrong here. The error is this:
```
julia> z45, delta45 = frule_via_ad(DiffractorRuleConfig(), (ZeroTangent(),1), sqrt, 2.0)
ERROR: MethodError: no method matching iterate(::Diffractor.TangentBundle{1, Float64, Tuple{Float64}})
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:737
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:737
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at dict.jl:695
  ...
Stacktrace:
 [1] indexed_iterate(I::Diffractor.TangentBundle{1, Float64, Tuple{Float64}}, i::Int64)
   @ Base ./tuple.jl:92
 [2] top-level scope
   @ REPL[54]:1
```
It certainly calls `shuffle_base(r) = TangentBundle{1}(r[1], (r[2],))` (10 lines up) but after that I don't know.

simeonschaub: I don't think you can destructure a Tangent Bundle like this.

mcabbott: But what is destructuring it? The thing I construct called `tangents` is just a Tuple, and looks acceptable to `∂☆internal{1}()`. From the type `tangents[2]` is being iterated but I don't see where.

simeonschaub: My suspicion is that it's you doing `z45, delta45 = frule_via_ad(...)` inside the test, but I could be wrong.

mcabbott: Oh right, ok. That would explain the short stack trace! This must return a tuple but surely I can figure out how to convert it.

simeonschaub:
CI failure is https://github.com/JuliaLang/julia/pull/42125, will push a fix shortly if no one beats me to it.

simeonschaub:
Just as a small comment: I'd generally always squash-merge PRs like this unless the commits are actually standalone changes.

mcabbott:
Sorry, will do, next time.

PR #47, opened 2021-09-10
update dependencies and add compat

simeonschaub:
- add compat entries
- update dependencies

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,275 +1,274 @@
 # This file is machine-generated - editing it directly is not advised
 
-[[Adapt]]
+manifest_format = "2.0"
+
+[[deps.Adapt]]
 deps = ["LinearAlgebra"]
 git-tree-sha1 = "84918055d15b3114ede17ac6a7182f68870c16f7"
 uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 version = "3.3.1"
 
-[[ArgTools]]
+[[deps.ArgTools]]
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -15,6 +15,10 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 [compat]
 ChainRules = "1.5"
 ChainRulesCore = "1.2"
+Combinatorics = "1"
+StaticArrays = "1"
+StatsBase = "0.33"
+StructArrays = "0.6"
 julia = "1"
```
mcabbott, line 22: Should this be 1.7?

simeonschaub: Yes, it probably should.
```suggestion
julia = "1.7"
```

Manifest.toml
```diff
@@ -1,275 +1,274 @@
 # This file is machine-generated - editing it directly is not advised
 
-[[Adapt]]
```
mcabbott, line 3: Not new, but is it a good idea to check in the manifest? 

Will this constrain CI to run these exact versions, or will it resolve to the latest allowed?

simeonschaub: I think CI will use these versions if possible. I believe the Manifest was originally checked in because Diffractor depended on different branches of some dependencies, but I don't think that's the case anymore. @keno, should we just delete the Manifest?

simeonschaub: I think I'll merge this as-is and then we can revisit whether we want to just delete the manifest later.

PR #50, opened 2021-09-10
set up some docs

simeonschaub:


simeonschaub:
The failure of the Documentation job here is https://github.com/JuliaDocs/Documenter.jl/issues/1693, so this should be fine once merged/

PR #54, opened 2021-09-13
WIP: Refactor forward mode data structures

Keno:
To allow chunking like ForwardDiff.

I think this works on the Diffractor side at this point, but there's a question about the interfaces with ChainRules. In particular, to what extent do we want to want/need ChainRules to be chunking-aware. Previously ChainRules did allow tuples/arrays in its forwards rules, but we removed that support in preparation for Diffractor, because having broadcasting in the rules makes them a lot more expensive at higher orders. I think our options here are:

1. Put the broadcasting back into ChainRules
2. Have some sort of `ImplicitBroadcast` type that wraps our values and implicitly broadcasts them as appropriate (so the rules look "scalar", but actually work with chunked values).
3. Option 2 + some compiler support to implement `ImplicitBroadcast` even if the type annotations are too restrictive.
4. Change the `frule` interface to also use closures like rrule, with the semantics being that for chunked evaluation, you broadcast the pushforward over the tangent vectors.

My current thinking is that either option 3 or 4 is perhaps attractive, but I'd like @oxinabox @mcabbott and @simeonschaub to weigh in with their opinions.

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -1,15 +1,14 @@
-partial(x::TangentBundle, i) = x.partials[i]
-partial(x::TaylorBundle{1}, i) = x.coeffs[i]
-partial(x::UniformBundle, i) = x.partial
-partial(x::CompositeBundle{N, B}, i) where {N, B} = Tangent{B}(map(x->partial(x, i), x.tup)...)
-partial(x::ZeroTangent, i) = ZeroTangent()
+partial(x::TangentBundle, i) = partial(getfield(x, :tangent), i)
+partial(x::ExplicitTangent, i) = getfield(getfield(x, :partials), i)
+partial(x::TaylorTangent, i) = getfield(getfield(x, :coeffs), i)
+partial(x::UniformTangent, i) = getfield(x, :val)
+partial(x::ProductTangent, i) = ProductTangent(map(x->partial(x, i), getfield(x, :factors)))
+partial(x::AbstractZero, i) = x
+partial(x::CompositeBundle{N, B}, i) where {N, B} = Tangent{B}(map(x->partial(x, i), getfield(x, :tup))...)
 primal(x::AbstractTangentBundle) = x.primal
 primal(z::ZeroTangent) = ZeroTangent()
 
-first_partial(x::TangentBundle{1}) = getfield(getfield(x, :partials), 1)
-first_partial(x::TaylorBundle{1}) = getfield(getfield(x, :coeffs), 1)
-first_partial(x::UniformBundle) = getfield(x, :partial)
-first_partial(x::CompositeBundle) = map(first_partial, getfield(x, :tup))
+first_partial(x) = partial(x, 1)
 
 # TODO: Which version do we want in ChainRules?
 function my_frule(args::ATB{1}...)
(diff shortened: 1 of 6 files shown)


oxinabox:
4 is a nonstarter.
We used to do that.
But we removed it because solving ODEs and their sensitivities (including chunked) is best done all at once in one DE solve step, that that's primal and sensibility inputs.
Search ChainRulesCore for fusing.

It's also breaking, so not doing that.
For some cases though you can use the new experimental feature for derivatives without augmented primals, but that is not available for all cases. (And in theory is available for only a smaller infinite number of functions)

2 sounds good to me

test/runtests.jl
```diff
@@ -210,4 +210,10 @@ z45, delta45 = frule_via_ad(DiffractorRuleConfig(), (0,1), x -> log(exp(x)), 2)
 @test z45 ≈ 2.0
 @test delta45 ≈ 1.0
 
+# Chunked Gradients
+using Diffractor: TangentBundle, ZeroBundle, ProductTangent, ExplicitTangent
```
simeonschaub, line 214:
```suggestion
using Diffractor: TangentBundle, ZeroBundle, ProductTangent, ExplicitTangent, ∂⃖¹
```

src/tangent.jl
```diff
@@ -78,33 +78,104 @@ function Base.getindex(a::AbstractTangentBundle, b::TaylorTangentIndex)
...
+
+    c₀ + c₁ ∂₁ + c₁ ∂₂ + c₁ ∂₃ + c₂ ∂₂ ∂₁ + c₂ ∂₃ ∂₁ + c₂ ∂₃ ∂₂ + c₃ ∂₃ ∂₂ ∂₁
+
+i.e.
+
+    c₀ + c₁ (∂₁ + ∂₂ + ∂₃) + c₂ (∂₂ ∂₁ + ∂₃ ∂₁ + ∂₃ ∂₂) + c₃ ∂₃ ∂₂ ∂₁
+
+
+This restriction forms a submanifold of the original manifold. The naming is
+by analogy with the (truncated) Taylor series
```
simeonschaub, line 112: I must admit, I am still a little confused about the name `TaylorTangent` and why it's useful. That's mostly unrelated to this PR though.

simeonschaub (commented):
I'd probably also be in favor of 2 or 3. I think it's probably best not to introduce another layer of closures into the `frule` convention.

Could you expand on what you mean by:

>     3\. Option 2 + some compiler support to implement `ImplicitBroadcast` even if the type annotations are too restrictive.

? In which cases is this necessary?

test/runtests.jl
Keno: Heh, this is actually a typo on my part. Should be the forward mode function of course. I had fixed that locally, but neglected to save the file.

src/tangent.jl
Keno: It implements Taylor mode AD like https://github.com/JuliaDiff/TaylorSeries.jl.

test/runtests.jl
simeonschaub: Ah, I see! It's difficult to tell them apart in a lot of fonts.

Keno: Yeah ... But it's so cool though.

simeonschaub: True

src/tangent.jl
simeonschaub: I thought it still had separate coefficients for each variable, or did I misunderstand what the `∂ₙ` are? https://github.com/JuliaDiff/TaylorSeries.jl#examples

Keno: This docstring shows the analogy between the `∂ₙ` and the coefficients of a taylor series. In particular, taylor mode takes advantage of symmetry of the derivative operators (i.e. the order of the operators is not distinguishable). For example, you can't use taylor mode to evaluate the second derivative in `∂₁ sin(∂₂ sin(x))`, but you can in `∂^2 sin(sin(x)) =  ∂₁ ∂₂ sin(sin(x)) = ∂₂ ∂₁`. Because of this symmetry under exchange of the operators, you also get symmetry in the tangent space that you can then represent explicitly.

src/stage1/forward.jl
```diff
@@ -1,14 +1,14 @@
...
 primal(x::AbstractTangentBundle) = x.primal
 primal(z::ZeroTangent) = ZeroTangent()
 
-first_partial(x::TangentBundle{1}) = getfield(getfield(x, :partials), 1)
-first_partial(x::TaylorBundle{1}) = getfield(getfield(x, :coeffs), 1)
-first_partial(x::UniformBundle) = getfield(x, :partial)
+first_partial(x::ExplicitTangentBundle{1}) = getfield(getfield(getfield(x, :tangent), :partials), 1)
+first_partial(x::TaylorBundle{1}) = getfield(getfield(getfield(x, :tangent), :coeffs), 1)
+first_partial(x::UniformBundle) = getfield(getfield(x, :tangent), :val)
 first_partial(x::CompositeBundle) = map(first_partial, getfield(x, :tup))
```
simeonschaub, line 12: Do we actually need to define `first_partial` separately? Can't we just use `partial(x, 1)` instead?

simeonschaub:
I am wondering whether it might even be enough to just overload `+` and scalar multiplication for `ProductTangent` (and perhaps some others ChainRulesCore defines for `AbstractTangent`s). Or would that quickly run into issues?

AnasAbdelR (approved):

AnasAbdelR:
I had this PR open out of pure interest, accidently hit approved thinking I was on my other tab, please disregard 😅

Keno:
I think I'm pretty convinced that the broadcast problem should be solved in a separate package that support SIMT semantics over arbitrary julia functions that we would depend on here.

Keno:
I'm planning to merge this as is and then do further work on this on master. Looks like there may still be some bugs remaining, but I want to merge the various threads first.

PR #55, opened 2021-09-15
bump compact bound to Julia 1.8

MasonProtter:
This package is using `Base.@constprop` in many places, which does not exist in Julia 1.7, rather it'll be a 1.8 feature if I understand correctly.


simeonschaub:
That change will be backported to 1.7

MasonProtter:
Hm. Do you know when that'll happen?

simeonschaub:
I thought it was already in RC1, but it needed to be backported manually. Should be in RC2.

Issue #70, opened 2022-01-12
Forward mode with complex numbers


mcabbott:
These should agree, right?
```julia
julia> Diffractor.PrimeDerivativeBack(x -> real(sin(x)))(1+im)
0.8337300251311491 + 0.9888977057628651im

julia> Diffractor.PrimeDerivativeFwd(x -> real(sin(x)))(1+im)
ERROR: Tangent space not defined for `Complex{Int64}.
```

PR #73, opened 2022-01-12
Take testing seriously

### LINKED ISSUE #67: Failure with ComposedFunction `∘`
I'm pretty confident this worked in September, but have no idea whether changes here or in ChainRules broke it:
```julia
julia> Diffractor.gradient(cbrt, 1.23)
(0.29036348772107673,)

julia> Diffractor.gradient(identity∘cbrt, 1.23)
ERROR: ArgumentError: Tangent for the primal Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}} should be backed by a AbstractDict type, not by NamedTuple{(:data,), Tuple{ChainRulesCore.ZeroTangent}}.
Stacktrace:
  [1] _backing_error(P::Type, G::Type, E::Type)
    @ ChainRulesCore ~/.julia/packages/ChainRulesCore/qzYOG/src/tangent_types/tangent.jl:62
  [2] ChainRulesCore.Tangent{Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, NamedTuple{(:data,), Tuple{ChainRulesCore.ZeroTangent}}}(backing::NamedTuple{(:data,), Tuple{ChainRulesCore.ZeroTangent}})
    @ ChainRulesCore ~/.julia/packages/ChainRulesCore/qzYOG/src/tangent_types/tangent.jl:33
  [3] (::Diffractor.var"#162#164"{Symbol, DataType})(Δ::ChainRulesCore.ZeroTangent)
    @ Diffractor ~/.julia/packages/Diffractor/HYuxt/src/stage1/generated.jl:308
  [4] (::Diffractor.EvenOddOdd{1, 1, Diffractor.var"#162#164"{Symbol, DataType}, Diffractor.var"#163#165"{Symbol}})(Δ::ChainRulesCore.ZeroTangent)
    @ Diffractor ~/.julia/packages/Diffractor/HYuxt/src/stage1/generated.jl:288
  [5] ∂⃖¹₁merge
    @ ./none:1
  [6] ∂⃖¹₁
    @ ./none:1
  [7] (::Diffractor.ApplyOdd{1, 1})(Δ::Float64)
    @ Diffractor ~/.julia/packages/Diffractor/HYuxt/src/stage1/generated.jl:371
  [8] ∂⃖¹₁ComposedFunction
    @ ./none:1
  [9] (::Diffractor.∇{ComposedFunction{typeof(identity), typeof(cbrt)}})(args::Float64)
    @ Diffractor ~/.julia/packages/Diffractor/HYuxt/src/interface.jl:122
 [10] Diffractor.∇(::Function, ::Float64)
    @ Diffractor ~/.julia/packages/Diffractor/HYuxt/src/interface.jl:128
 [11] top-level scope
    @ REPL[3]:1
```

mcabbott:
This adds many tests from Zygote. I've done some filtering to leave out the most repetitive ones, things which clearly don't apply (like Params), and things testing rules for external packages.

All which didn't pass are marked broken... ~~with (the local version of) #68. Thus many will fail.~~ Some have the error copied next to them. Exactly two have SILENTLY WRONG ANSWER in the comment (and should probably become issues) but this is not automatically distinguished from just an error.

~~Probably we should move the existing tests here into another file or two, to be a bit more organised.~~ (done) ~~I also had somewhere a few test cases of rules in ChainRules which are meant to support higher derivatives, but not very well tested there to do so.~~ (some added)

Closes #67, by adding the minimal piracy needed to allow Pairs to have a NamedTuple gradient (since many others tests failed without this).


Keno:
ForwardDiff and TaylorSeries are also sources of tests since we have and want to explicitly support forward mode.

mcabbott:
Good point. Ideally as many as possible of these would run all ways. The extent of my attempt at this so far is [these lines](https://github.com/JuliaDiff/Diffractor.jl/pull/73/files#diff-fba4d4a7c64aacd9b1a662bca4487f3e849b43f09e65f6c8788a3f142b989f65R20-R28) , which ran into https://github.com/JuliaDiff/Diffractor.jl/issues/70 .

mcabbott:
These tests really want a different `@test_broken` macro, which accepts an error as a broken test, but fails on `false` -- a silently wrong answer is much worse than an error. Has anyone got one in their desk drawer?

ToucheSir:
I looked into this a while back for an unrelated issue, but it seems the functionality behind those macros is not terribly extensible. Perhaps worthy of creating a TestMacros or TestExtras package if one doesn't already exist?

oscardssmith:
should this be updated and merged?

mcabbott:
Will circle back, but last time I tried to update this, so many things were broken that I stopped.

oscardssmith:
makes sense. Right now first order scalar forward mode is probably the thing that is most reliable.

staticfloat:
> These tests really want a different @test_broken macro, which accepts an error as a broken test, but fails on false -- a silently wrong answer is much worse than an error. Has anyone got one in their desk drawer?

For what it's worth, this was fixed in https://github.com/JuliaLang/julia/pull/47804

mcabbott:
Thanks. But this isn't quite what's wanted here, which is to prioritise silently wrong answers. It does this:
```julia
julia> @test_broken error()  # this I want to accept -- broken functionality
Test Broken
  Expression: error()

julia> @test_broken false  # this I would like to fail -- it's a silently wrong gradient
Test Broken
  Expression: false

julia> @test_broken 1  # this changed in 47804, ok
Error During Test at REPL[832]:1
  Expression evaluated to non-Boolean
```

staticfloat:
You don't want `@test_broken` then, you want `@test_throws`:

```
julia> @test_throws ErrorException error()
Test Passed
      Thrown: ErrorException

julia> @test_throws ErrorException false
Test Failed at REPL[5]:1
  Expression: false
    Expected: ErrorException
  No exception thrown
ERROR: There was an error during testing

julia> @test_throws ErrorException 1
Test Failed at REPL[6]:1
  Expression: 1
    Expected: ErrorException
  No exception thrown
ERROR: There was an error during testing
```

mcabbott:
Maybe. My hope was that an upgrade which changes an error to a correct answer would cause warnings (to fix at leisure) not failure, but changing an error to false would cause warnings (as now you have silently wrong answers). Maybe describing that as a variant of `@test_broken` is weird, since "Unexpected Pass" is a failure not a warning.

Issue #69, opened 2022-01-11
Accumulation


mcabbott:
At present Diffractor may return thunks, but doesn't seem use them (or anything else) to accumulate effciently:
```julia
julia> @btime gradient(x -> sum(x), $(rand(100, 100))) |> first |> typeof
  min 1.829 μs, mean 1.869 μs (8 allocations, 368 bytes)
ChainRulesCore.InplaceableThunk{ChainRulesCore.Thunk...

julia> @btime gradient(x -> sum(x) + sum(x), $(rand(100, 100))) |> first |> typeof
  min 8.917 μs, mean 28.032 μs (22 allocations, 235.03 KiB)
Matrix{Float64} (alias for Array{Float64, 2})

julia> @btime copy($(rand(100, 100)));
  min 1.262 μs, mean 4.120 μs (2 allocations, 78.17 KiB)

julia> 235.03 / 78.17
3.00665216835103
```
Should this change, perhaps to use `ChainRulesCore.add!!`? In which case it might it be easiest to change now, while there is nothing downstream to break.

Or do we count on `ImmutableArray` and compiler improvements?

Xref https://github.com/FluxML/Zygote.jl/pull/981 which retrofits Zygote to accumulate in-place.

PR #79, opened 2022-07-15
always unthunk results

oscardssmith:
in theory this shouldn't be necessary, but it is a good fallback to make sure we don't return thunks to the user. Also, it's 0 overhead as long as the preceding calculation is correctly inferred.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -159,7 +159,7 @@ function (f::PrimeDerivativeBack)(x)
     z = ∂⃖¹(lower_pd(f), x)
     y = getfield(z, 1)
     f☆ = getfield(z, 2)
-    return getfield(f☆(dx(y)), 2)
+    return unthunk(getfield(f☆(dx(y)), 2))
 end
 
 # Forwards primal derivative
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -127,7 +127,7 @@ end
 # N.B: This means the gradient is not available for zero-arg function, but such
 # a gradient would be guaranteed to be `()`, which is a bit of a useless thing
 function (::Type{∇})(f, x1, args...)
-    ∇(f)(x1, args...)
+    unthunk.(∇(f)(x1, args...))
 end
 
 const gradient = ∇
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -108,7 +108,7 @@ let var"'" = Diffractor.PrimeDerivativeBack
     # Control flow cases
     @test @inferred((x->simple_control_flow(true, x))'(1.0)) == sin'(1.0)
     @test @inferred((x->simple_control_flow(false, x))'(1.0)) == cos'(1.0)
-    @test_broken (x->sum(isa_control_flow(Matrix{Float64}, x)))'(Float32[1 2;]) == [1.0 1.0;]
+    @test (x->sum(isa_control_flow(Matrix{Float64}, x)))'(Float32[1 2;]) == [1.0 1.0;]
     @test times_three_while'(1.0) == 3.0
 
     pow5p(x) = (x->mypow(x, 5))'(x)


mcabbott:
This is intended to only affect the outermost level, the thing returned to the user who types `f'(x)`, right?

Should the same be done to `gradient(f, xs...)`?

oscardssmith:
this should be done to gradient. diffractor would actually prefer if thunks didn't exist in the first place, so this is just a fallback to prevent the user from seeing them. our ideal scenario would be a way to get derivatives from chainrules that don't have thunks in the first place, but that is a lot harder than this pr.

mcabbott:
Some way of not computing things which will be discarded does seem desirable. With the example of  https://github.com/JuliaDiff/ChainRulesCore.jl/issues/558 (but picturing `a,b` huge arrays):
```julia
julia> Diffractor.PrimeDerivativeBack(x -> f(x, b) + 10f(2x, b))(a)
rrule is called
rrule is called
∇a is called
∇a is called
42.0f0  # with this PR, seems ideal? ∇b never run despite accumulation

julia> gradient(x -> f(x, b), a)  # should unthunk answer, but not run ∇b
rrule is called
(Thunk(var"#28#31"{Float32, Float32, Float32}(1.0f0, 1.0f0, 2.0f0)),)
```

oscardssmith:
The goal for diffractor is to use escape analysis to remove the computation entirely which is made easier with simpler types. We aren't there yet, but once stage 2 is integrated, we'll be closeish.

mcabbott:
This would be great. If it works, it's possible all thunks could be stripped out of ChainRules, since (IIRC) nothing else uses them anyway?

Although, besides delayed calculation, they are intended one day to save memory too... xref #69 I guess.

oscardssmith:
as I understand it, ReverseDiff.jl likes thunks, but I'm not sure.

mcabbott:
Yes I suppose, although you have to opt-in. More directly I forgot that Yota doesn't un-thunk internally, only the final result: 
```julia
julia> using Yota

julia> grad(x -> f(x, b), a)  # unthunks final result
rrule is called
∇a is called
(2.0f0, (ZeroTangent(), 2.0f0))

julia> using Zygote

julia> Zygote.gradient(x -> f(x, b), a)  # unthunks all
rrule is called
∇a is called
∇b is called
(2.0f0,)

julia> using ReverseDiff

julia> ReverseDiff.@grad_from_chainrules f(x::TrackedReal, y::Real)

julia> ReverseDiff.gradient(x -> f(x[1], b), [a])  # maybe this relies on thunks inside?
rrule is called
∇a is called
1-element Vector{Float32}:
 2.0
```

src/interface.jl
```diff
@@ -159,9 +162,21 @@ function (f::PrimeDerivativeBack)(x)
...
+end
+
+
+struct JacobianBack{N, T} <: AbstractPrimeDerivative{N, T}
+    f::T
+end
+
+function (f::JacobianBack)(x)
+    y, f☆ = ∂⃖(getfield(f, :f), x)
+    return transpose(reduce(hcat, f☆(collect(row))[2] for row in dx_jacobian(y)))
```
mcabbott, line 175: I think some `vec` may be missing:
```
julia> Diffractor.jacobian(x -> x[1:2])(rand(3,3))
6×3 transpose(::Matrix{Float64}) with eltype Float64:
 1.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> Zygote.jacobian(x -> x[1:2], rand(3,3))[1]
2×9 Matrix{Float64}:
 1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```
And maybe a `tuple`?
```
julia> Diffractor.gradient(atan)(1/2)
(0.8,)

julia> Diffractor.gradient(atan)(1, 2)
(0.4, -0.2)

julia> Diffractor.jacobian(x -> x[1:2])([3,4,5])  # shouldn't this also be a tuple?
2×3 transpose(::Matrix{Int64}) with eltype Int64:
 1  0  0
 0  1  0

julia> Diffractor.jacobian(*)(rand(2,2), rand(2))  # to allow this...
ERROR: MethodError: no method matching (::Diffractor.JacobianBack{1, typeof(*)})(::Matrix{Float64}, ::Vector{Float64})
```

mcabbott: Or maybe jacobian deserves to be another PR. What should these do?
```
julia> Diffractor.PrimeDerivativeBack(x -> sum(sum(x.a) .+ x.b))((a=[1,2], b=[3,4], c=[5,6]))  # maybe this should unthunk within Tangent? 
Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}}(b = [1.0, 1.0], a = InplaceableThunk(ChainRules.var"#..., Thunk(ChainRules.var"#...)))

julia> Diffractor.gradient(x -> sum(sum(x.a) .+ x.b), (a=[1,2], b=[3,4], c=[5,6]))[1]  # ditto?
Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}}(b = [1.0, 1.0], a = InplaceableThunk(ChainRules.var"#..., Thunk(ChainRules.var"#...)))

julia> Diffractor.jacobian(x -> sum(x.a) .+ x.b)((a=[1,2], b=[3,4], c=[5,6]))  # what structure should this return?
2×1 transpose(::Matrix{Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}, NamedTuple{(:b, :a), Tuple{Vector{Int64}, InplaceableThunk{Thunk{ChainRules.var"#1573#1576"{Int64, Colon, Vector{Int64}, ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}, ChainRules.var"#1572#1575"{Int64, Colon}}}}}}) with eltype Union{}:
Error showing value of type Transpose{Union{}, Matrix{Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}, NamedTuple{(:b, :a), Tuple{Vector{Int64}, InplaceableThunk{Thunk{ChainRules.var"#1573#1576"{Int64, Colon, Vector{Int64}, ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}, ChainRules.var"#1572#1575"{Int64, Colon}}}}}}}:
ERROR: MethodError: no method matching transpose(::Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}, NamedTuple{(:b, :a), Tuple{Vector{Int64}, InplaceableThunk{Thunk{ChainRules.var"#1573#1576"{Int64, Colon, Vector{Int64}, ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}, ChainRules.var"#1572#1575"{Int64, Colon}}}}})
```

oscardssmith: honestly I meant to make it a separate PR and forgot to check out a new PR.

mcabbott: Can this branch `unthunk` apply to `gradient` too, and perhaps within `Tangent`?

oscardssmith:
I think this is good to merge. (jacobians have been extricated).

mcabbott:
Can this branch unthunk apply to gradient too, and perhaps within Tangent? Examples from above:
```
julia> Diffractor.PrimeDerivativeBack(x -> sum(sum(x.a) .+ x.b))((a=[1,2], b=[3,4], c=[5,6]))  # maybe this should unthunk within Tangent? 
Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}}(b = [1.0, 1.0], a = InplaceableThunk(ChainRules.var"#..., Thunk(ChainRules.var"#...)))

julia> Diffractor.gradient(x -> sum(sum(x.a) .+ x.b), (a=[1,2], b=[3,4], c=[5,6]))[1]  # ditto?
Tangent{NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}}}(b = [1.0, 1.0], a = InplaceableThunk(ChainRules.var"#..., Thunk(ChainRules.var"#...)))
```

oscardssmith:
I think this should be merged as is. I haven't found a great way to `unthunk` within `Tangent`, and these are more of a fallback anyway until we find a better way to get `rrule`s that don't have `Thunk`s in the first place.

mcabbott (approved):
Seems ok by me. Would be nice if tests passed, though...

For disabling thunks, xref https://github.com/JuliaDiff/ChainRulesCore.jl/pull/568 (inspired by difficulties with Zygote over Zygote).

mcabbott:
Master has the same errors now, on Julia nightly: `((((sin')')')')(1.0) == sin(1.0)  MethodError: no method matching fieldnames(::Nothing)`, so it's not this PR's fault.

PR #81, opened 2022-07-19
Add jacobian

oscardssmith:
This uses a slightly different definition of the jacobian than ForwardDiff. For jacobian(f:N×M->P) it will return a result of shape `P×N×M` while ForwardDiff will return `P×(N⋅M)`


mcabbott:
Shouldn't this return a tuple, like gradient, to allow for multiple arguments? 
```
julia> Diffractor.gradient(atan)(1/2)
(0.8,)

julia> Diffractor.gradient(atan)(1, 2)
(0.4, -0.2)

julia> Diffractor.jacobian(x -> x[1:2])([3,4,5])
2×3 transpose(::Matrix{Int64}) with eltype Int64:
 1  0  0
 0  1  0

julia> Diffractor.jacobian(*)(rand(2,2), rand(2)) 
ERROR: MethodError: no method matching (::Diffractor.JacobianBack{1, typeof(*)})(
```
... and allow `jacobian(f, x...)` do allow `do` blocks?
```
julia> Diffractor.jacobian() do x
         x[1:2]
       end([1,2,3])
2×3 Matrix{Int64}:
 1  0  0
 0  1  0
```
One argument against this make-more-dimensions story is that few other things work like that. Not just all the other AD packages, but for instance `*` does not:
```
julia> rand(2)' * Diffractor.jacobian(x -> x[1:2])(rand(4))
1×4 adjoint(::Vector{Float64}) with eltype Float64:
 0.036297  0.265575  0.0  0.0

julia> rand(2)' * Diffractor.jacobian(x -> x[1:2])(rand(3,4))  # makes a 2×3×4 Array{Float64, 3}
ERROR: MethodError: no method matching *(::Adjoint{Float64, Vector{Float64}}, ::Array{Float64, 3})
```
If it only accepts arrays of as arguments, then perhaps this should be in its signature, rather than trying & failing. Perhaps only arrays of numbers in fact:
```
julia> Diffractor.jacobian(x -> x[1])([[1,2], [3,4]])
ERROR: MethodError: no method matching zero(::Type{Vector{Int64}})

julia> Diffractor.jacobian(x -> [x, 2x])(3.0)
ERROR: MethodError: no method matching reshape(::Float64, ::Tuple{Int64})

julia> Diffractor.jacobian(x -> sum(x.a) .+ x.b)((a=[1,2], b=[3,4], c=[5,6])) 
ERROR: MethodError: no method matching size(::NamedTuple{(:a, :b, :c), Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}})
```

PR #82, opened 2022-08-01
Support getindex on non-numeric arrays

ToucheSir:
This came up while trying to run Flux models with `Chain{<:Vector}`.


mcabbott:
This case should work with https://github.com/JuliaDiff/ChainRules.jl/pull/655, but Diffractor's own rule takes precedence.

PR #86, opened 2022-08-18
Fix `rrule_via_ad` return type

mcabbott:
This ought to return a tuple; what it returns now works in some contexts but not https://github.com/JuliaDiff/ChainRules.jl/pull/644 . 
```julia
julia> using Diffractor, ChainRulesCore

julia> ret = rrule_via_ad(Diffractor.DiffractorRuleConfig(), x -> log(exp(x)), 2)
Diffractor.OpticBundle{Float64}(2.0, (::Any)::Tuple{ZeroTangent, Any}->◌)

julia> y, bk = ret;  # ok

julia> Tuple(ret)
ERROR: MethodError: no method matching length(::Diffractor.OpticBundle{Float64})
```
Originally added in https://github.com/JuliaDiff/Diffractor.jl/pull/45, thus my fault...

### DIFF
--- a/src/stage1/generated.jl
+++ b/src/stage1/generated.jl
@@ -227,7 +227,7 @@ function (::∂⃖{N})(f::T, args...) where {T, N}
 end
 
 function ChainRulesCore.rrule_via_ad(::DiffractorRuleConfig, f::T, args...) where {T}
-    ∂⃖{1}()(f, args...)
+    ∂⃖{1}()(f, args...) |> Tuple{<:Any, <:Any}
 end
 
 @Base.pure function (::∂⃖{1})(::typeof(Core.apply_type), head, args...)
--- a/src/stage1/generated.jl
+++ b/src/stage1/generated.jl
@@ -227,7 +227,7 @@ function (::∂⃖{N})(f::T, args...) where {T, N}
 end
 
 function ChainRulesCore.rrule_via_ad(::DiffractorRuleConfig, f::T, args...) where {T}
-    ∂⃖{1}()(f, args...) |> Tuple{<:Any, <:Any}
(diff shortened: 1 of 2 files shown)


src/stage1/generated.jl
```diff
@@ -227,7 +227,7 @@ function (::∂⃖{N})(f::T, args...) where {T, N}
 end
 
 function ChainRulesCore.rrule_via_ad(::DiffractorRuleConfig, f::T, args...) where {T}
-    ∂⃖{1}()(f, args...)
+    ∂⃖{1}()(f, args...) |> Tuple{<:Any, <:Any}
```
JeffBezanson, line 230:
```suggestion
    ∂⃖{1}()(f, args...) |> Tuple{Any, Any}
```

JeffBezanson:
That `length` method for `OpticBundle` should be added.

src/stage1/generated.jl
mcabbott: Sorry!

mcabbott:
Is it ok to merge this?

src/stage1/generated.jl
```diff
@@ -227,7 +228,7 @@ function (::∂⃖{N})(f::T, args...) where {T, N}
 end
 
 function ChainRulesCore.rrule_via_ad(::DiffractorRuleConfig, f::T, args...) where {T}
-    ∂⃖{1}()(f, args...)
+    ∂⃖{1}()(f, args...) |> Tuple{Any, Any}
```
mcabbott, line 231: Slightly nicer:
```suggestion
    Tuple{Any, Any}(∂⃖{1}()(f, args...))
```

PR #88, opened 2022-08-31
Accumulate NamedTuple + Tangent

mcabbott:
This is a hack to fix problems like this:
```julia
g_kw(;x=1.0) = sin(x)
f_kw(x) = g_kw(;x)
@test bwd(f_kw)(1.0) == bwd(sin)(1.0)  broken=true
#=
MethodError: no method matching +(::Tangent{var"#g_kw#47"{var"#g_kw#11#48"}, NamedTuple{(Symbol("#g_kw#11"),), Tuple{ZeroTangent}}}, ::Tangent{Diffractor.KwFunc{var"#g_kw#47"{var"#g_kw#11#48"}, var"#g_kw#47##kw"}, NamedTuple{(:kwf,), Tuple{ZeroTangent}}})
...
  [2] elementwise_add(a::NamedTuple{(:contents,), Tuple{Tangent{var"#g_kw#47"{var"#g_kw#11#48"}, NamedTuple{(Symbol("#g_kw#11"),), Tuple{ZeroTangent}}}}}, b::NamedTuple{(:contents,), Tuple{Tangent{Diffractor.KwFunc{var"#g_kw#47"{var"#g_kw#11#48"}, var"#g_kw#47##kw"}, NamedTuple{(:kwf,), Tuple{ZeroTangent}}}}})
    @ ChainRulesCore ~/.julia/packages/ChainRulesCore/ctmSK/src/tangent_types/tangent.jl:287
  [3] +(a::Tangent{Core.Box, NamedTuple{(:contents,), Tuple{Tangent{var"#g_kw#47"{var"#g_kw#11#48"}, NamedTuple{(Symbol("#g_kw#11"),), Tuple{ZeroTangent}}}}}}, b::Tangent{Core.Box, NamedTuple{(:contents,), Tuple{Tangent{Diffractor.KwFunc{var"#g_kw#47"{var"#g_kw#11#48"}, var"#g_kw#47##kw"}, NamedTuple{(:kwf,), Tuple{ZeroTangent}}}}}})
    @ ChainRulesCore ~/.julia/packages/ChainRulesCore/ctmSK/src/tangent_arithmetic.jl:130
=#
```
Diffractor sometimes produces ChainRulesCore's Tangent types, and sometimes produces plain NamedTuples instead. If these happen for structural gradients of the same object, then you will get an error. 

Ideally it should probably make up its mind. Maybe `Tangent`s are a pain to deal with at higher order?

### DIFF
--- a/src/extra_rules.jl
+++ b/src/extra_rules.jl
@@ -266,3 +266,6 @@ end
 function ChainRulesCore.rrule(::DiffractorRuleConfig, ::Type{InplaceableThunk}, add!!, val)
     val, Δ->(NoTangent(), NoTangent(), Δ)
 end
+
+Base.real(z::ZeroTangent) = z  # TODO should be in CRC
+Base.real(z::NoTangent) = z
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -13,3 +13,17 @@ end
 @Base.constprop :aggressive accum(a::NoTangent, b) = b
 @Base.constprop :aggressive accum(a, b::NoTangent) = a
 @Base.constprop :aggressive accum(a::NoTangent, b::NoTangent) = NoTangent()
+
+using ChainRulesCore: Tangent, backing
+
+function accum(x::Tangent{T}, y::NamedTuple) where T
+  # @warn "gradient is both a Tangent and a NamedTuple" x y
+  z = accum(backing(x), y)
+  Tangent{T,typeof(z)}(z)
+end
+accum(x::NamedTuple, y::Tangent) = accum(y, x)
+
+function accum(x::Tangent{T}, y::Tangent) where T
+  z = accum(backing(x), backing(y))
+  Tangent{T,typeof(z)}(z)
+end
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -5,25 +5,25 @@ struct DiffractorRuleConfig <: RuleConfig{Union{HasReverseMode,HasForwardsMode}}
 @Base.constprop :aggressive accum(a::Tuple, b::Tuple) = map(accum, a, b)
 @Base.constprop :aggressive @generated function accum(x::NamedTuple, y::NamedTuple)
     fnames = union(fieldnames(x), fieldnames(y))
+    isempty(fnames) && return :((;))  # code below makes () instead
     gradx(f) = f in fieldnames(x) ? :(getfield(x, $(quot(f)))) : :(ZeroTangent())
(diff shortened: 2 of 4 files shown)


src/runtime.jl
```diff
@@ -5,11 +5,25 @@ struct DiffractorRuleConfig <: RuleConfig{Union{HasReverseMode,HasForwardsMode}}
 @Base.constprop :aggressive accum(a::Tuple, b::Tuple) = map(accum, a, b)
```
mcabbott, line 5: I get some errors from this adding tuples of mis-matched lengths, which I haven't tracked down. Not touched in this PR.

What I don't know here is whether `accum` is only ever called on a single tangent, or also on a tuple of a function's arguments. IIRC there is no distinction in Zygote, but in CR the former can be Array or Tangent or NoTangent, but the latter can only ever be a Tuple (of the right length).

src/extra_rules.jl
```diff
@@ -266,3 +266,6 @@ end
 function ChainRulesCore.rrule(::DiffractorRuleConfig, ::Type{InplaceableThunk}, add!!, val)
     val, Δ->(NoTangent(), NoTangent(), Δ)
 end
+
+Base.real(z::ZeroTangent) = z  # TODO should be in CRC
+Base.real(z::NoTangent) = z
```
mcabbott, line 271: This is https://github.com/JuliaDiff/ChainRulesCore.jl/pull/581 now

oscardssmith:
I don't think this is the right long term solution, but it seems right for now.

mcabbott:
Yes for sure it's a hack. But it doesn't seem like it will hide serious bugs, and having tests that work for now is good.

Last commit disables tests on 1.8, since judging by "┌ Warning: ir verification broken. Either use 1.9 or 1.7" there isn't an intention to make everything work there. (The actual failures are inference tests.)

ToucheSir:
On the topic of longer-term solutions, why is `Box` appearing here at all? kwsorters are usually type stable and Diffractor doesn't appear to change their branching in ways which would disrupt that. `Diffractor.KwFunc` also doesn't make use of any inner closures.

mcabbott:
I don't know. Where should we put things like this, where a failing test has uncovered bad behaviour, but possibly no longer bad enough to make the test fail? 

(Fun Zygote example where re-using a variable name causes a 5x slowdown: [Z#1290](https://github.com/FluxML/Zygote.jl/issues/1290#issuecomment-1221404923))

PR #92, opened 2022-09-29
RFC: run optimization passes at the end of `codegen`

JeffBezanson:


staticfloat:
@Keno what's the status of this?  Should we close this?

Keno:
Yes, close it. We'll need to redo it anyway on top of current APIs.

PR #94, opened 2022-12-27
Diffractor: Bump to julia 1.10, prepare for stage 2 work

Keno:
This:
- Bumps the Diffractor version to 0.2.0 (in preparation, will be tagged once new things are in).
- Bumps the julia dependency version to 1.10
- Gets rid of conditional code for older julia code.

This is in preparation for the stage2 code that will depend on lattice extensions and other 1.10-only features. If there is interest, we can branch off 2e571fe50409b6b6e9c3d7bb662c68cca666dfd4 as release-0.1, but if nobody complains, I'd rather save the effort to focus on current development.


Keno:
Depends on https://github.com/JuliaLang/julia/pull/48005 to avoid spamming errors on nightly.

PR #107, opened 2023-02-11
updates to work on Julia master

oscardssmith:
https://github.com/JuliaLang/julia/pull/48601 added a new `VarState` struct to represent undef static params.

PR #108, opened 2023-02-11
removes `@pure` and some miscelanious cleanups

oscardssmith:
this is NFC, just removing a lot of uses of `@pure` that @keno isn't allowed to write since he isn't Jamison (and they were probably from before the effect system).


oscardssmith:
CI doesn't pass on nightly without https://github.com/JuliaDiff/Diffractor.jl/pull/107, but it passes locally.

oscardssmith:
I think this is ready to merge.

Keno:
CI is failing.

PR #109, opened 2023-02-11
Fearless taylor version

oscardssmith:
If we're willing to use the information that derivatives are linear operators, we can remove a lot of complexity from forward mode differentiation (and improve type stability in the process). I know @Keno doesn't believe this is valid to assume, but IMO the simplification enabled provides a pretty strong justification that this is the right approach.


Keno (changes requested):
This is unsound as is, even if you want to make the assumption that taylor-ness is preserved. This just makes everything taylor which is unconditionally unsound.

oscardssmith:
in that case we definitely need a test case that breaks because locally all the tests pass with this PR. (also, can you decipher the error message on CI I'm not able to understand or reproduce it).

PR #110, opened 2023-02-14
add handling for lshr_int

pepijndevos:
I'm not sure if this is needed or correct but the bsim_group tran tests errors here (after warning about incidence though, so maybe this would have just been inlined)


Keno:
This is indicating some missing rule that should intercept this at a higher level. I don't think we should be adding hard coded rules to the intrinsics.

PR #114, opened 2023-03-07
adjust to the latest JuliaLang/julia#master

aviatesk:


src/codegen/forward_demand.jl
```diff
@@ -129,7 +129,7 @@ function forward_diff_uncached!(ir::IRCode, interp, irsv::IRInterpretationState,
         # Incidence analysis through the rt call
         # TODO: frule_mi is wrong here, should be the mi of the caller
         frule_rt = info.frule_call.rt
-        improve_frule_rt = CC.concrete_eval_invoke(interp, frule_call, frule_mi, irsv)
+        (improve_frule_rt, nothrow) = CC.concrete_eval_invoke(interp, frule_call, frule_mi, irsv)
```
oscardssmith, line 132: can you put this behind a version check?

aviatesk: Isn't Diffractor supposed to work only with Julia nightly?

oscardssmith: I thought it was supposed to be 1.8 or up, but @Keno should confirm.

Keno: No, we're nightly only

PR #121, opened 2023-03-21
Document and clean sig on forward_diff_no_inf

oxinabox:
I wanted to add a docstring as I had no idea what `visit_custom!` did.
And then i noticed that most of the other arguments are not used

### DIFF
--- a/src/codegen/forward_demand.jl
+++ b/src/codegen/forward_demand.jl
@@ -288,7 +288,6 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue, Int}};
     end
 end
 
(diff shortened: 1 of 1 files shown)


src/codegen/forward_demand.jl
```diff
@@ -271,11 +286,12 @@ function forward_diff_no_inf!(ir::IRCode, interp, mi::MethodInstance, world, to_
             end
         end
     end
-
 end
 
+Base.@deprecate forward_diff_no_inf!(ir::IRCode, interp, mi::MethodInstance, world, to_diff::Vector{Pair{SSAValue, Int}}; kwargs...) forward_diff_no_inf!(ir, to_diff; kwargs...)
```
oxinabox, line 291: Do we need this?
Given the package is currently unregistered.
And this is an internal method.

OTOH i know we have some internal code that calls this method directly

Keno: Not required

Keno (approved):

src/codegen/forward_demand.jl
oxinabox:
```suggestion
```

PR #137, opened 2023-04-09
Fix codegen of fwd varargs

Keno:


Keno:
Rebased and made the nesting actually work.

oxinabox:
I don't understand how you fixed varargs.

PR #142, opened 2023-04-17
testing if CI is broken

oscardssmith:
I believe CI is currently broken on julia master.


oxinabox:
reproduces on my local machine also

aviatesk:
So do you mean it is related to world age?

oxinabox:
I suspect so. But the MethodError message doesn't say it exists in a newer world. So perhaps not

PR #148, opened 2023-05-17
Custom forward rule for `Expr(:new, )` in `forward_diff_no_inf!`

staticfloat:
Frames and I tracked down an issue where the type argument to an `Expr(:new, )` (argument #1) was getting wrapped in a `ZeroBundle`. This commit adds a special-case for `Expr(:new, )` into `forward_diff_no_inf!()` that mimics what we see elsewhere, making use of `∂☆new`.


staticfloat:
@oxinabox can you make test for this please?

src/codegen/forward_demand.jl
```diff
@@ -286,6 +286,9 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
                 end
                 inst[:inst] = Expr(:call, ∂☆{order}(), newargs...)
                 inst[:type] = Any
+            elseif isexpr(stmt, :new)
+                inst[:inst] = Expr(:call, ∂☆new{order}(), stmt.args...)
```
Keno, line 290: I think you'll need to copy the `newargs` code above or merge the two cases.

Keno (approved):

test/stage2_fwd.jl
```diff
@@ -43,4 +43,23 @@ module stage2_fwd
...
+        struct Bar148
+            v
+        end
+        foo_148(x) = Bar148(x)
+
+        # this is needed as transform! is *always* called on Arguments regardless of what visit_custom says
+        identity_transform!(ir, ssa::Core.SSAValue, order) = ir[ssa]
+        function identity_transform!(ir, arg::Core.Argument, order)
+            return Core.Compiler.insert_node!(ir, Core.SSAValue(1), Core.Compiler.NewInstruction(Expr(:call, Diffractor.ZeroBundle{1}, arg), Any))
+        end
```
oxinabox, line 57: @Keno  it seems that this is always required.

PR #154, opened 2023-05-22
remove :loopinfo from IR

oscardssmith:
fixes `Diffractor.PrimeDerivativeFwd(x->prod([x, 4]))(3)`. (with help from @oxinabox)

### DIFF
--- a/src/codegen/forward.jl
+++ b/src/codegen/forward.jl
@@ -53,11 +53,12 @@ function fwd_transform!(ci, mi, nargs, N)
             return ZeroBundle{N}(mi.sparam_vals[stmt.args[1]::Int])
         elseif isexpr(stmt, :foreigncall)
             return Expr(:call, error, "Attempted to AD a foreigncall. Missing rule?")
-        elseif isexpr(stmt, :meta) || isexpr(stmt, :inbounds)
+        elseif isexpr(stmt, :meta) || isexpr(stmt, :inbounds)  || isexpr(stmt, :loopinfo)
             # Can't trust that meta annotations are still valid in the AD'd
             # version.
             return nothing
         else
+            #TODO put guard here. We really don't want invalid IR
             return Expr(:call, ZeroBundle{N}, stmt)
         end
     end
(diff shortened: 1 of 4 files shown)


src/codegen/forward_demand.jl
```diff
@@ -185,6 +185,8 @@ function forward_visit!(ir::IRCode, ssa::SSAValue, order::Int, ssa_orders::Vecto
         recurse(stmt)
     elseif isexpr(stmt, :code_coverage_effect)
         return
+    elseif isexpr(stmt, :loopinfo) #TODO preserve this properly
```
Keno, line 188: It's not legal to preserve.

src/codegen/forward.jl
```diff
@@ -53,11 +53,12 @@ function fwd_transform!(ci, mi, nargs, N)
             return ZeroBundle{N}(mi.sparam_vals[stmt.args[1]::Int])
         elseif isexpr(stmt, :foreigncall)
             return Expr(:call, error, "Attempted to AD a foreigncall. Missing rule?")
-        elseif isexpr(stmt, :meta) || isexpr(stmt, :inbounds)
+        elseif isexpr(stmt, :meta) || isexpr(stmt, :inbounds)  || isexpr(stmt, :loopinfo)
             # Can't trust that meta annotations are still valid in the AD'd
             # version.
             return nothing
         else
+            #TODO put guard here. We really don't want invalid IR
```
Keno, line 61: This is not in general invalid IR

src/codegen/forward_demand.jl
oxinabox: It can't be outright preserved but we might be able to do *something* with this infromation

src/codegen/forward.jl
oxinabox: The proposal here is rather that we assert that this is only getting various kinds of literals to wrap in a `ZeroBundle` (be they functions or numbers etc)
and if not give a good error message, rather than generating invalid IR.

PR #155, opened 2023-05-22
fix show `Type{<:AbstractZeroBundle}`

oscardssmith:
We arguably should delete this method for all the reasons that you never are supposed to define these, but if we are defining it, we should at least define it right.


oxinabox:
When I added it.  I did note that you were not supposed to define these. But honestly the stacktraces are *so* bad without it. 
I think we can have a little evil as a treat

PR #158, opened 2023-05-30
Give `Diffractor` a `ForwardDiff` inspired interface

oscardssmith:
This is still somewhat WIP, but the goal is to make an exported interface for users to use. I'm currently running into some issues since this changes `gradient` to be forward mode (since forward mode is generally a lot more reliable currently).


oxinabox:
I think the better way to do this is to extend [AbstractDifferentiation.jl](https://github.com/JuliaDiff/AbstractDifferentiation.jl/tree/master)
which exports names like this.

I mean this is a partway step towards that, so we could do this first then that, but idk the motivation goes down once the short term need is met.
And we would like to have it working with AbstractDifferentiation eventually anyway.

This does mean doing  `derivative(DiffractorFwdBackend(), f, x)` rather than `Diffractor.derivative(f, x)` but that seems fine.
And it gives us trivial swap-out for other ADs.
And gives us other things like lazy jacobian matrix-like operators (ForwardDiff2 style)

staticfloat:
> This does mean doing `derivative(DiffractorFwdBackend(), f, x)` rather than `Diffractor.derivative(f, x)` but that seems fine.

So can we just define `derivative(::DiffractorFwdBackend, f, x) = Diffractor.derivative(f, x)` and call it a day?

oxinabox:
We should indeed be able to do something like that yes.

Basically we need to define one of these 4 functions
https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/master/src/AbstractDifferentiation.jl#L495-L502

For example https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/041d760d4c0bff8a5e023f9d1d5cad2ffbb68f4d/ext/AbstractDifferentiationForwardDiffExt.jl#L28

One can define more to include optimized versions of stuff like `jacobian` (which we will want to do at some stage.). But it is not required.

oscardssmith:
Closing in favor of https://github.com/JuliaDiff/Diffractor.jl/pull/187

PR #162, opened 2023-06-14
Don't insert truncation within phi blocks

oxinabox:
This goes along with making diffractor follow this rule:
https://github.com/JuliaLang/julia/pull/50158

The hard bit of this PR was writing the test. But now I know kinda how phi-nodes work.

### DIFF
--- a/src/stage1/compiler_utils.jl
+++ b/src/stage1/compiler_utils.jl
@@ -64,3 +64,41 @@ Base.lastindex(x::Core.Compiler.InstructionStream) =
 if isdefined(Core.Compiler, :CallInfo)
     Base.convert(::Type{Core.Compiler.CallInfo}, ::Nothing) = Core.Compiler.NoCallInfo()
 end
+
+
+"""
+    find_end_of_phi_block(ir, start_search_idx)
+
+Finds the last index within the same basic block, on or after the `start_search_idx` which is not within a phi block.
+A phi-block is a run on PhiNodes or nothings that must be the first statements within the basic block.
+
+If `start_search_idx` is not within a phi block to begin with, then just returns `start_search_idx`
+"""
+function find_end_of_phi_block(ir, start_search_idx)
+    # Short-cut for early exit:
+    stmt = ir.stmts[start_search_idx][:inst]
+    stmt !== nothing && !isa(stmt, PhiNode) && return start_search_idx
+
+    # Actually going to have to go digging throught the IR to out if were are in a phi block
+    # TODO: this is not so efficient. maybe preconstruct CFG then use block_for_inst?
+    start_bb=0
+    found_arg = false
+    end_idx = start_search_idx + 1  # This goes one over, but we will substract one at the end
+    for (bb, idx) in bbidxiter(ir)
+        if !found_arg
+            if idx==start_search_idx
+                start_bb = bb
+                found_arg = true
+            end
+        else # found it
+            end_idx = idx  # This goes one over, but we will substract one at the end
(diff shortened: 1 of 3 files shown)


src/stage1/compiler_utils.jl
```diff
@@ -64,3 +64,41 @@ Base.lastindex(x::Core.Compiler.InstructionStream) =
...
+
+If `start_search_idx` is not within a phi block to begin with, then just returns `start_search_idx`
+"""
+function find_end_of_phi_block(ir, start_search_idx)
+    # Short-cut for early exit:
+    stmt = ir.stmts[start_search_idx][:inst]
+    stmt !== nothing && !isa(stmt, PhiNode) && return start_search_idx
+
+    # Actually going to have to go digging throught the IR to out if were are in a phi block
+    # TODO: this is not so efficient. maybe preconstruct CFG then use block_for_inst?
```
Keno, line 83: `block_for_inst(ir.cfg, idx)`

src/stage1/compiler_utils.jl
```diff
@@ -64,3 +64,41 @@ Base.lastindex(x::Core.Compiler.InstructionStream) =
...
+            # hit end of block, so saft to insert
+            bb !== start_bb && break
+            stmt = ir.stmts[idx][:inst]
+            # No longer in a phi block, so safe to insert
+            stmt !== nothing && !isa(stmt, PhiNode) && break
+        end
+    end
+    end_idx-=1
+    return end_idx
+end
```
oxinabox, line 104: This is not very efficient since it just does a linear search over whole IR, but I am not sure if it is worth it to practively construct CFG so can do a searchsorted. (Does bbidxiter generates a CFG under the hood?)

src/stage1/compiler_utils.jl
oxinabox: Oh the ir object has a CFG?

Keno:
Idea looks right. Algorithm for check the bb needs to be updated as suggested, otherwise looks fine.

src/stage1/compiler_utils.jl
```diff
@@ -64,3 +64,29 @@ Base.lastindex(x::Core.Compiler.InstructionStream) =
...
+function find_end_of_phi_block(ir, start_search_idx::Int)
+    # Short-cut for early exit:
+    stmt = ir.stmts[start_search_idx][:inst]
+    stmt !== nothing && !isa(stmt, PhiNode) && return start_search_idx
+
+    # Actually going to have to go digging throught the IR to out if were are in a phi block
+    # TODO: this is not so efficient. maybe preconstruct CFG then use block_for_inst?
+    bb=CC.block_for_inst(ir.cfg, start_search_idx)
+    end_search_idx=ir.cfg.blocks[bb].stmts[end]
+    for idx in (start_search_idx+1):(end_search_idx-1)
```
Keno, line 86: This is possibly off by one? What if `start_search_idx` is the last PhiNode?

Keno (approved):
LGTM, modulo comment

src/stage1/compiler_utils.jl
oxinabox: Ah yes, I thought we had already checked that one, but because this search is checking `idx+1` we haven't.
```suggestion
    for idx in (start_search_idx):(end_search_idx-1)
```

oxinabox:
Elliot bisected CI being broken to https://github.com/JuliaLang/julia/commit/9d839f9794cbe4cc5eb654bcf8064f3f30452705

PR #165, opened 2023-06-15
reorganize tests

oxinabox:
Just realised that `main` is still failing tests so expect this to fail.
This just moves tests around


oxinabox:
Ok, all test are passing, weird.
I guess the problem is I was running an old version of julia nightly locally.

PR #167, opened 2023-06-19
error for nonliterals

oxinabox:
I believe this is the correct way to detect literals. This has hit us a few times with some unexpected IR statment

### DIFF
--- a/src/codegen/forward.jl
+++ b/src/codegen/forward.jl
@@ -59,7 +59,7 @@ function fwd_transform!(ci, mi, nargs, N)
             return nothing
         else
             # Fallback case, for literals.
-            # Literals are not Exprs so error for them.
+            # If it is an Expr, then it is not a literal
             if isa(stmt, Expr)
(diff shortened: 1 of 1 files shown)


oxinabox:
@oscardssmith  this was a todo we wrote together, do you want to review?

src/codegen/forward.jl
```diff
@@ -58,8 +58,11 @@ function fwd_transform!(ci, mi, nargs, N)
             # version.
             return nothing
         else
-            #TODO put guard here. We really don't want to generate invalid IR
-            # by wrapping something that isn't more or less a literal in a ZeroBundle
+            # Fallback case, for literals.
+            # Literals are not Exprs so error for them.
```
oxinabox, line 62:
```suggestion
            # If it is an Expr, then it is not a literal
```

staticfloat (approved):
If this passes with DAECompiler, I say merge it.  It should be a nice way to quickly find issues with statements we're not expecting.

PR #168, opened 2023-06-27
Improve type inference of non-`const` GlobalRef's

staticfloat:
GlobalRefs often do not get their types refined, so we should be careful to re-use types that we have when possible.  This PR moves non-`const` GlobalRef handling out of `maparg()` so that we can keep the original type.

### DIFF
--- a/src/codegen/forward_demand.jl
+++ b/src/codegen/forward_demand.jl
@@ -256,11 +256,7 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
             # TODO: Should we remember whether the callbacks wanted the arg?
             return transform!(ir, arg, order)
         elseif isa(arg, GlobalRef)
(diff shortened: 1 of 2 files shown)


oscardssmith:
Needs a test, but otherwise looks good.

staticfloat:
Test added, luckily it was not as difficult as I feared.  I have verified that the test fails without this change.  Unfortunately, the reverse mode tests were broken by https://github.com/JuliaLang/julia/commit/ec3319430bd88a32b588b09491aa9fd94bf75733 and that might be a heavy lift to fix.  We'll check in with Keno tomorrow to see how bad it's going to be to fix that (he says it's a miscompile, which sounds bad).  If it's too too painful we can either disable those reverse-mode tests (since we're not using them anyway), look into some kind of workaround, or even merge this with the reverse mode tests failing.

oxinabox (approved):

PR #169, opened 2023-06-27
Pass `ssa` instead of `stmt` to `visit_custom!()`

staticfloat:
We need to pass the SSA index, rather than the statement itself so that `visit_custom!()` can inspect the type of the statement under visitation.

This is more consistent with the `transform!()` API as well.


oxinabox:
This is a good change.

There are several tests which hack around this lack.
In the `forward_diff_no_inf` tests 


They should be updated to not work around it.

oxinabox:
Hmm this did not fix those tests.  No matter, I will follow up on that later.

Issue #170, opened 2023-06-28
Reverse mode tests broken


oxinabox:
The reverse mode tests were broken by https://github.com/JuliaLang/julia/commit/ec3319430bd88a32b588b09491aa9fd94bf75733 and that might be a heavy lift to fix.  We'll check in with Keno tomorrow to see how bad it's going to be to fix that (he says it's a miscompile, which sounds bad).  If it's too too painful we can either disable those reverse-mode tests (since we're not using them anyway), look into some kind of workaround, or even merge this with the reverse mode tests failing.

_Originally posted by @staticfloat in https://github.com/JuliaDiff/Diffractor.jl/issues/168#issuecomment-1610414663_

PR #171, opened 2023-06-28
Mark broken reverse mode tests broken

oxinabox:
Work around #170 
This marks over 70 tests broken.
Once that is resolved can find and replace to remove the markers. (or revert this commit)
I will merge this ASAP so can merge other stuff

### DIFF
--- a/test/regression.jl
+++ b/test/regression.jl
@@ -13,18 +13,18 @@ const bwd = Diffractor.PrimeDerivativeBack
 
 
 # Regression tests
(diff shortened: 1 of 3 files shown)


oxinabox:
I do not get the failure related to effect-free on my local machine with yesterday's nightly.
Rebuidling my nightly now to investigate

aviatesk (approved):
Fine to merge this quickly since it looks like we can remove these `broken` annotations easily using a grep replacement once we resolve the issue.

PR #172, opened 2023-06-28
remove noneffectfree per JuliaLang/julia#50313

oxinabox:
This is follow up to  JuliaLang/julia#50313

 - We no longer import things that are not defined
 - We do import the new `effect_free_and_nothrow` but we don't actually use it (this seemed like correct match to use previously importing `effect_free` and not using it
 - we stop wrapping an instruction in `non_effect_free`. Because that is no longer defined and a similar change was made to an instruction in  JuliaLang/julia#50313 so i think it is ok, but i am not sure why it was there in the first place

### DIFF
--- a/src/stage1/recurse.jl
+++ b/src/stage1/recurse.jl
@@ -3,8 +3,8 @@ using Core.Compiler:
     Instruction, MethodInstance, NewInstruction, NewvarNode, OldSSAValue, PhiNode,
     ReturnNode, SSAValue, SlotNumber, StmtRange,
     bbidxiter, cfg_delete_edge!, cfg_insert_edge!, compute_basic_blocks, complete,
-    construct_domtree, construct_ssa!, domsort_ssa!, effect_free, finish, insert_node!,
-    insert_node_here!, non_dce_finish!, non_effect_free, quoted, retrieve_code_info,
+    construct_domtree, construct_ssa!, domsort_ssa!, finish, insert_node!,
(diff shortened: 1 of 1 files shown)


oxinabox:
Remaining CI failures are fixed by #171

aviatesk (approved):

aviatesk:
> we stop wrapping an instruction in non_effect_free. Because that is no longer defined and a similar change was made to an instruction in Align meaning for effects and IR flags JuliaLang/julia#50313 so i think it is ok, but i am not sure why it was there in the first place

We recompute statement flags of `NewInstruction` if it is not given explicitly. `non_effect_free` utility was used as the alias to `NewInstruction(stmt, ..., flag=recompute_stmt_flags(stmt) & ~IR_FLAG_EFFECT_FREE)` but it turned out it's not that useful or necessary. For our case `recompute_stmt_flag` won't give `IR_FLAG_EFFECT_FREE` for our special statement anyway.

aviatesk:
> we stop wrapping an instruction in non_effect_free. Because that is no longer defined and a similar change was made to an instruction in Align meaning for effects and IR flags JuliaLang/julia#50313 so i think it is ok, but i am not sure why it was there in the first place

We recompute statement flags of `NewInstruction` if it is not given explicitly. `non_effect_free` utility was used as the alias to `NewInstruction(stmt, ..., flag=recompute_stmt_flags(stmt) & ~IR_FLAG_EFFECT_FREE)` but it turned out it's not that useful or necessary. For our case `recompute_stmt_flag` won't give `IR_FLAG_EFFECT_FREE` for our special statement anyway.

PR #173, opened 2023-06-28
bipass all Diffractor machinery if there are no partials

oxinabox:
> 🎵 I mean, what have you got to lose?
> You know, you come from nothing
> You're going back to nothing
> What have you lost? Nothing
> Nothing will come from nothing, ya know what they say 🎵

Related to the recent work by  @Keno  and @staticfloat.

By the linearity of the derivative operation if the inputs are all zero so is the output.
A case can be made that we actually should be still invoking the `frule`s for this path.
But I am not sure that's a good argument. 
Here the derivative isn't just `iszero` it has been put into the types it is zero and can't ever not be zero.
(This would not for example be the case if it were being mutated)

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -137,6 +137,18 @@ function (::∂☆shuffle{N})(args::AbstractTangentBundle{N}...) where {N}
     ∂☆p(ZeroBundle{N-1}(frule), #= ZeroBundle{N-1}(DiffractorRuleConfig()), =# tupargs, map(primal, downargs)...)
 end
 
+# Special shortcut case if there is no derivative information at all:
+function (::∂☆internal{N})(f::AbstractZeroBundle{N}, args::AbstractZeroBundle{N}...) where {N}
+    f_v = primal(f)
+    args_v = map(primal, args)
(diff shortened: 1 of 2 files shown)


src/stage1/forward.jl
```diff
@@ -137,6 +137,18 @@ function (::∂☆shuffle{N})(args::AbstractTangentBundle{N}...) where {N}
...
+function (::∂☆internal{N})(f::AbstractZeroBundle{N}, args::AbstractZeroBundle{N}...) where {N}
+    f_v = primal(f)
+    args_v = map(primal, args)
+    return ZeroBundle{N}(f_v(args_v...))
+end
+function (::∂☆internal{1})(f::AbstractZeroBundle{1}, args::AbstractZeroBundle{1}...)
+    f_v = primal(f)
+    args_v = map(primal, args)
+    return ZeroBundle{1}(f_v(args_v...))
+end
```
staticfloat, line 150: Why do you have a special case for order 1?

oxinabox: Break method ambiguity

oscardssmith (approved):

staticfloat:
I think this makes sense, so I'm just going to merge it.

PR #177, opened 2023-06-29
Ensure that `:boundscheck` is properly handled

staticfloat:
My previous fix was incorrect because `:boundscheck` returns a `Bool`. Let's just force it to `true` for now, which implicitly disables `@inbounds`, which is probably a good idea for us since we haven't gone through the effort of ensuring that all indexing is still valid. Add tests for this so that we catch this breakage faster next time.

### DIFF
--- a/test/forward.jl
+++ b/test/forward.jl
@@ -69,4 +69,25 @@ end
     @test primal_calls[] == 1
 end
 
+@testset "indexing" begin
+    # Test to make sure that `:boundscheck` and such are properly handled
+    function foo(x)
+        t = (x, x)
+        return t[1] + 1
+    end
+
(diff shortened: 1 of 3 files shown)


test/forward.jl
```diff
@@ -69,4 +69,25 @@ end
...
+    end
+
+    # Test that `@inbounds` is ignored by Diffractor
+    function foo_errors(x)
+        t = (x, x)
+        @inbounds return t[3] + 1
+    end
+    let var"'" = Diffractor.PrimeDerivativeFwd
+        @test_throws BoundsError foo_errors'(1.0) == 1.0
+    end
```
oscardssmith, line 90: you might want to add a test of a function with `@inbounds` that has a chain rule with invalid indexing.

PR #179, opened 2023-07-08
Allow visit_custom recursion to change the order

Keno:
To implement order-changing intrinsics.

### DIFF
--- a/src/tangent.jl
+++ b/src/tangent.jl
@@ -283,7 +283,7 @@ end
 Base.getindex(u::UniformBundle, ::TaylorTangentIndex) = u.tangent.val
 
 """
-    CompositeBundle{N, B <: Tuple}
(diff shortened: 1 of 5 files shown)


staticfloat:
@keno can you add a simple test that changes the order?

src/tangent.jl
```diff
@@ -283,7 +283,7 @@ end
 Base.getindex(u::UniformBundle, ::TaylorTangentIndex) = u.tangent.val
 
 """
-    CompositeBundle{N, B <: Tuple}
+    CompositeBundle{N, B, B <: Tuple}
```
oxinabox, line 286:
```suggestion
    CompositeBundle{N, B, T <: Tuple}
```

oxinabox (approved):
This makes sense.
One comment that is unrelated to the bulk of this

PR #180, opened 2023-07-10
Use `replace_call!()` to replace `Expr(:call, ...)` values

staticfloat:
This is necessary to prevent the callinfo field from falling out of sync with the call itself, causing future optimization passes (such as inlining) to compute incorrect results.


staticfloat:
@oxinabox Can I give you the task of finding other places within Diffractor where we replace `Expr(:call, ..)` values, and use `replace_call!()` there as well?  We don't have to do this for IR statements we're inserting, only ones we're overwriting.

oxinabox:
sure, will do

oxinabox:
Having looked at every instance of `Expr(:call` and `:inst]` and `stmt` I can't find any others
there is one that replaces it in a `IncrementalCompact` but i think that is fine.
the rest are inserts.

PR #181, opened 2023-07-11
[NFC] Insert truncation/primal just before statement

oxinabox:
NFC = No functional change

This is a more elegant way to handle this than https://github.com/JuliaDiff/Diffractor.jl/pull/162
we just insert the function that is applied to the arguments just before the statement, rather than just after the argument.
thus avoiding need to dodge phi nodes, and making the ir code more readable.
This is also consistent with how we handle `GlobalRef`s

### DIFF
--- a/src/codegen/forward_demand.jl
+++ b/src/codegen/forward_demand.jl
@@ -241,9 +241,9 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
                 @assert order < argorder
                 return get!(truncation_map, arg=>order) do
                     if order == 0
-                        insert_node!(ir, arg, NewInstruction(Expr(:call, primal, arg), Any), #=attach_after=#true)
+                        insert_node!(ir, ssa, NewInstruction(Expr(:call, primal, arg), Any), #=attach_after=#false)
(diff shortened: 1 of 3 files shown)


Keno (changes requested):
I don't think this is necessarily correct. It's certainly not NFC

oxinabox:
Isn't it?
By no functional change I mean, it should have no user visible effects.
But it's not a simply code cleanup that is true. It does change the IR generated.
But not in a way that should effect what actually ends up running.
But if I broke it then obviously it would.

What does it get incorrect?

for some input
```
%1 = a
%2 = b
%3 = c
%4 = f(%1)
```
Before this PR we would get
```
%1 = a
%2 = primal(%1)
%3 = b
%4 = c
%5 = f(%2)
```
After this PR we would get 
```
%1 = a
%2 = b
%3 = c
%4 = primal(%1)
%4 = f(%4)
```

Keno:
What happens if the `ssa` is a PhiNode? What prevents us from trying to insert a call in the middle of the phi block?

oxinabox:
ooh, I never considered that. I was only thinking of `invoke`s and `call`s.
Yeah, this approach won't work out.

PR #182, opened 2023-07-13
stage2(forward): add method table backedge for non-existing `frule` method

aviatesk:


staticfloat (approved):

staticfloat:
It appears to me that this is working as expected, the other failures seem unrelated to me.

aviatesk:
Yes, the failures reproduce on the main branch too. So they should be fixed separately.

PR #185, opened 2023-07-17
minor test improvements

oscardssmith:
These are a few fixes to CI that I found while trying to give Diffractor an interface.

### DIFF
--- a/src/codegen/reverse.jl
+++ b/src/codegen/reverse.jl
@@ -465,7 +465,8 @@ function diffract_ir!(ir, ci, meth, sparams::Core.SimpleVector, nargs::Int, N::I
             elseif isexpr(stmt, :static_parameter)
                 fwds[i] = ZeroTangent()
             elseif isa(stmt, Union{GotoNode, GotoIfNot})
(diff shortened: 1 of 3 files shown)


src/codegen/reverse.jl
```diff
@@ -465,7 +465,8 @@ function diffract_ir!(ir, ci, meth, sparams::Core.SimpleVector, nargs::Int, N::I
             elseif isexpr(stmt, :static_parameter)
                 fwds[i] = ZeroTangent()
             elseif isa(stmt, Union{GotoNode, GotoIfNot})
-                return :(error("Control flow support not fully implemented yet for higher-order reverse mode (TODO)"))
+                #@show stmt
+                error("Control flow support not fully implemented yet for higher-order reverse mode (TODO)")
```
oxinabox, line 469: Why do we want to move this error to compile time,  rather then runtime?

oscardssmith: well if we want it at run time, we need to generate ir rather than an expression

oxinabox: oh true

oxinabox (approved):

oscardssmith:
re-running CI now that Julia master is fixed.

PR #190, opened 2023-07-25
stop depending on StatsBase

oxinabox:
We don't seem to use it


oxinabox:
we were using it for `countmap` but that ius trivial to implement, so i did

PR #203, opened 2023-08-14
remove getindex definition on 1.11

oscardssmith:
We recently defined this in Base.


oscardssmith:
@Keno do you understand why macos CI is failing but linux ci is working?

PR #205, opened 2023-08-21
add more types supported with bundle helper

oxinabox:


oxinabox:
sigh, this was slightly different to what I was doing when i tested it out,
and i introduced ambiguity errors.
I have a fix. Which I will put in soon

PR #215, opened 2023-09-08
add truncate for explicit tangent

pepijndevos:


oxinabox:
This seems right to me, can we just add some unit tests?
like just directly calling the method.

pepijndevos:
done

oxinabox (approved):

Issue #189, opened 2023-07-25
Support mutation for forward mode


oxinabox:
We currently have partial support for mutation in forwards mode, (I think)
I believe mutating arrays works, but mutating structs does not.

We need two things for this:
 - [ ] https://github.com/JuliaDiff/ChainRulesCore.jl/issues/105 a MutableTangent type
 - [ ] To never automatically insert `ZeroTangent` for mutable structs, and instead us a `MutableTangent` with no fields/all fields set to ZeroTangents (unless those fields themselves are mutable structs)
 
Mutation in forwards mode is much easier than reverse.

PR #216, opened 2023-09-19
Remove Composite Bundle

oxinabox:
This is a prerequisite for  #189  as how CompositeBundle works,  causes a problem when it comes to mutation.

Right now `CompositeBundle` stores everything as a tuple of other Bundles, one per field of the primal.
Rather than storing things as a a a primal and a  tangent (or really 1 or more tangents).
This is a problem because when we need to get stuff back seperated into primal and tangent (which we need to do to apply `frules` or to actually get at the derivative result) we need to assemble the Tangent (note the capital T, this is the type that represents an immutable structural tangent)  on the fly.
Which we do here: https://github.com/JuliaDiff/Diffractor.jl/blob/ee8af85b04da82a3292cc071df9059d87d1d00fa/src/stage1/forward.jl#L8-L14
That is all well and good if everything is immuatable, since there is no need to preserve mutation or worry about aliasing if things are immuable.
If on the other hand, things are  mutable, you do need to worry about this things. In particular the `MutableTangent` type must be aliased in the same places its primal is (infact in the PR https://github.com/JuliaDiff/ChainRulesCore.jl/pull/626/ that created that, I always made a note in the docstring), you very much can not break it a part and recreate it on demand.
Ergo, I think CompositeBundle needs to be removed and replaced with something that works like the other tangent types, a pair of primal and tangent values.
Namely `TaylorTangent` in most cases, though in theory it might end up with an `ExplictTangent`

However, actually removing `CompositeBundle` seems a net significant simplification of the code base.
This PR is net  negative in lines of code in `src/` and will get even more negative once I work out which of the remaining const prop aggressive special cases can be deleted.
It also fixes some existing issues and I think will make others easier to fix in the future.

### DIFF
--- a/src/stage1/compiler_utils.jl
+++ b/src/stage1/compiler_utils.jl
@@ -7,7 +7,9 @@ function Base.push!(cfg::CFG, bb::BasicBlock)
     push!(cfg.index, bb.stmts.start)
 end
 
-Base.getindex(ir::IRCode, ssa::SSAValue) = Core.Compiler.getindex(ir, ssa)
+if VERSION < v"1.11.0-DEV.258"
+    Base.getindex(ir::IRCode, ssa::SSAValue) = Core.Compiler.getindex(ir, ssa)
+end
 
 Base.copy(ir::IRCode) = Core.Compiler.copy(ir)
 
--- a/src/stage1/compiler_utils.jl
+++ b/src/stage1/compiler_utils.jl
@@ -7,9 +7,7 @@ function Base.push!(cfg::CFG, bb::BasicBlock)
     push!(cfg.index, bb.stmts.start)
 end
 
-if VERSION <= v"1.11.0-DEV.116"
-    Base.getindex(ir::IRCode, ssa::SSAValue) = Core.Compiler.getindex(ir, ssa)
-end
+Base.getindex(ir::IRCode, ssa::SSAValue) = Core.Compiler.getindex(ir, ssa)
 
 Base.copy(ir::IRCode) = Core.Compiler.copy(ir)
 
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -4,14 +4,6 @@ partial(x::TaylorTangent, i) = getfield(getfield(x, :coeffs), i)
 partial(x::UniformTangent, i) = getfield(x, :val)
 partial(x::ProductTangent, i) = ProductTangent(map(x->partial(x, i), getfield(x, :factors)))
 partial(x::AbstractZero, i) = x
-partial(x::CompositeBundle{N, B}, i) where {N, B<:Tuple} = Tangent{B}(map(x->partial(x, i), getfield(x, :tup))...)
-function partial(x::CompositeBundle{N, B}, i) where {N, B}
-    # This is tangent for a struct, but fields partials are each stored in a plain tuple
-    # so we add the names back using the primal `B`
-    # TODO: If required this can be done as a `@generated` function so it is type-stable
-    backing = NamedTuple{fieldnames(B)}(map(x->partial(x, i), getfield(x, :tup)))
-    return Tangent{B, typeof(backing)}(backing)
-end
 
 
 primal(x::AbstractTangentBundle) = x.primal
@@ -42,14 +34,6 @@ function shuffle_down(b::TaylorBundle{N, B}) where {N, B}
         ntuple(_sdown, N-1))
 end
 
-function shuffle_down(b::CompositeBundle{N, B}) where {N, B}
-    z = CompositeBundle{N-1, CompositeBundle{1, B}}(
-        (CompositeBundle{N-1, Tuple}(
-            map(shuffle_down, b.tup)
-        ),)
-    )
-    z
-end
 
 function shuffle_up(r::TaylorBundle{1, Tuple{B1,B2}}) where {B1,B2}
     z₀ = primal(r)[1]
@@ -63,18 +47,7 @@ function shuffle_up(r::TaylorBundle{1, Tuple{B1,B2}}) where {B1,B2}
     end
 end
 
-function shuffle_up(r::CompositeBundle{1})
-    z₀ = primal(r.tup[1])
-    z₁ = partial(r.tup[1], 1)
-    z₂ = primal(r.tup[2])
-    z₁₂ = partial(r.tup[2], 1)
-    if z₁ == z₂
-        return TaylorBundle{2}(z₀, (z₁, z₁₂))
-    else
-        return ExplicitTangentBundle{2}(z₀, (z₁, z₂, z₁₂))
-    end
-end
-
+#==
 function taylor_compatible(a::ATB{N}, b::ATB{N}) where {N}
     primal(b) === a[TaylorTangentIndex(1)] || return false
     return all(1:(N-1)) do i
@@ -88,7 +61,7 @@ isswifty(::UniformBundle) = true
 isswifty(b::CompositeBundle) = all(isswifty, b.tup)
 isswifty(::Any) = false
 
-#TODO: port this to TaylorTangent:
+#TODO: port this to TaylorTangent over composite structures
 function shuffle_up(r::CompositeBundle{N}) where {N}
     a, b = r.tup
     if isswifty(a) && isswifty(b) && taylor_compatible(a, b)
@@ -102,6 +75,7 @@ function shuffle_up(r::CompositeBundle{N}) where {N}
             ntuple(i->partial(b,i), 1<<(N+1)-1)...))
     end
 end
+==#
 
 function shuffle_up(r::UniformBundle{N, B, U}) where {N, B, U}
     (a, b) = primal(r)
@@ -198,13 +172,6 @@ end
         map(y->lifted_getfield(y, s), x.tangent.coeffs))
 end
 
-@Base.constprop :aggressive function (::∂☆{N})(::ATB{N, typeof(getfield)}, x::CompositeBundle{N}, s::AbstractTangentBundle{N, Int}) where {N}
-    x.tup[primal(s)]
-end
-
-@Base.constprop :aggressive function (::∂☆{N})(::ATB{N, typeof(getfield)}, x::CompositeBundle{N, B}, s::AbstractTangentBundle{N, Symbol}) where {N, B}
-    x.tup[Base.fieldindex(B, primal(s))]
-end
 
 @Base.constprop :aggressive function (::∂☆{N})(f::ATB{N, typeof(getfield)}, x::UniformBundle{N, <:Any, U}, s::AbstractTangentBundle{N}) where {N, U}
     UniformBundle{N,<:Any,U}(getfield(primal(x), primal(s)), x.tangent.val)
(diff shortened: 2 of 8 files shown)


src/stage1/forward.jl
```diff
@@ -89,6 +75,7 @@ function shuffle_up(r::CompositeBundle{N}) where {N}
             ntuple(i->partial(b,i), 1<<(N+1)-1)...))
     end
 end
+==#
```
oxinabox, line 78: Porting this over requires some thinking but shouldn't be too hard

src/stage1/forward.jl
```diff
@@ -185,13 +172,6 @@ end
         map(y->lifted_getfield(y, s), x.tangent.coeffs))
 end
 
-@Base.constprop :aggressive function (::∂☆{N})(::ATB{N, typeof(getfield)}, x::CompositeBundle{N}, s::AbstractTangentBundle{N, Int}) where {N}
-    x.tup[primal(s)]
-end
-
-@Base.constprop :aggressive function (::∂☆{N})(::ATB{N, typeof(getfield)}, x::CompositeBundle{N, B}, s::AbstractTangentBundle{N, Symbol}) where {N, B}
-    x.tup[Base.fieldindex(B, primal(s))]
-end
```
oxinabox, line 194: this shouldn't be needed as we already have cases for `TaylorBundle` above.
Likewise several others of this kind below will not be needed.

test/forward.jl
```diff
@@ -40,6 +39,7 @@ let var"'" = Diffractor.PrimeDerivativeFwd
 end
 
 # Some Basic Mixed Mode tests
+# TODO: unbreak this
```
oxinabox, line 42: idk why this is broken

test/forward.jl
```diff
@@ -90,4 +90,46 @@ end
...
+    end
+
+    function foo(a)
+        obj = IDemo(2.0, a)
+        return obj.x * obj.y
+    end
+
+    let var"'" = Diffractor.PrimeDerivativeFwd
+        @test foo'(100.0) == 2.0
+        @test foo''(100.0) == 0.0
```
oxinabox, line 107: this second derivative is actually broken on main

src/tangent.jl
```diff
@@ -202,15 +202,13 @@ end
...
 function TaylorBundle{N, B}(primal::B, coeffs) where {N, B}
-#    check_taylor_invariants(coeffs, primal, N) # TODO: renable this
+    check_taylor_invariants(coeffs, primal, N)
     _TangentBundle(Val{N}(), primal, TaylorTangent(coeffs))
 end
 
 function check_taylor_invariants(coeffs, primal, N)
     @assert length(coeffs) == N
-    if isa(primal, TangentBundle)
-        @assert isa(coeffs[1], TangentBundle)
```
oxinabox, line 212: this check is just wrong AFAICT.
because if the primal is a TangentBundle, then the first derivative is a `Tangent{TangentBundle}`
The first deriviative has no primal component.

But this was never hit before because we were not making `TaylorBundles` when taking derivatives of TaylorBundles, (at least not in anything that happens in our `test/forward.jl` I checked)  we were (I assume) making `CompositeBundles`

src/stage1/forward.jl
```diff
@@ -70,26 +54,33 @@ function taylor_compatible(a::ATB{N}, b::ATB{N}) where {N}
...
-        return TangentBundle{N+1}(r.tup[1].primal,
-            (r.tup[1].tangent.partials..., primal(b),
-            ntuple(i->partial(b,i), 1<<(N+1)-1)...))
+        #XXX: am dubious of the correctness of this
+        a_partials = ntuple(i->partial(r, ii)[1], N)
+        b_partials = ntuple(i->partial(r, ii)[2], N)
+        the_partials = (a_partials..., primal_b, b_partials...)
+        return TangentBundle{N+1}(the_primal, the_partials)
     end
 end
```
oxinabox, line 81: For reference as a part way step to determining this , I first refactored the CompositeBundle version into

```julia
function shuffle_up(r::CompositeBundle{N}) where {N}
    a, b = r.tup
    if isswifty(a) && isswifty(b) && taylor_compatible(a, b)
        the_partials = ntuple(N+1) do i
            if ii <= N
                a[TaylorTangentIndex(i)]  # == b[TaylorTangentIndex(i-1)] (except first which is b.primal)
            else  # ii = N+1
                b[TaylorTangentIndex(i-1)]                
            end
        end
        return TaylorBundle{N+1}(primal(a), the_partials)
    else
        the_primal = r.tup[1].primal
        a_partials = r.tup[1].tangent.partials
        b_partials = ntuple(i->partial(b,i), 1<<(N+1)-1) 
        the_partials = (a_partials..., primal_b, b_partials...)
        return TangentBundle{N+1}(the_primal, the_partials)
    end
end
```

src/tangent.jl
```diff
@@ -230,6 +228,18 @@ function Base.getindex(tb::TaylorBundle, tti::CanonicalTangentIndex)
...
+"for a TaylorTangent{N, <:Tuple} this breaks it up unto 1 TaylorTangent{N} for each element of the primal tuple"
+function destructure(r::TaylorBundle{N, B}) where {N, B<:Tuple}
+    return ntuple(fieldcount(B)) do field_ii
+        the_primal = primal(r)[field_ii]
+        the_partials = ntuple(N) do  order_ii
+            partial(r, order_ii)[field_ii]
+        end
+        return TaylorBundle{N}(the_primal, the_partials)
+    end
+end
```
oxinabox, line 240: Do we need this, or can we just use `unbundle`?

src/stage1/compiler_utils.jl
```diff
@@ -7,9 +7,7 @@ function Base.push!(cfg::CFG, bb::BasicBlock)
     push!(cfg.index, bb.stmts.start)
 end
 
-if VERSION <= v"1.11.0-DEV.116"
-    Base.getindex(ir::IRCode, ssa::SSAValue) = Core.Compiler.getindex(ir, ssa)
-end
```
oxinabox, line 12: It just doesn't work on my version of 1.11 without defining this.
idk what was up with 1.11.0DEV.116, maybe a change that was since reverted?

src/stage1/forward.jl
```diff
@@ -70,26 +54,33 @@ function taylor_compatible(a::ATB{N}, b::ATB{N}) where {N}
...
+        return TaylorBundle{N+1}(the_primal, the_partials)
     else
-        return TangentBundle{N+1}(r.tup[1].primal,
-            (r.tup[1].tangent.partials..., primal(b),
-            ntuple(i->partial(b,i), 1<<(N+1)-1)...))
+        #XXX: am dubious of the correctness of this
+        a_partials = ntuple(i->partial(r, ii)[1], N)
+        b_partials = ntuple(i->partial(r, ii)[2], N)
+        the_partials = (a_partials..., primal_b, b_partials...)
+        return TangentBundle{N+1}(the_primal, the_partials)
```
oxinabox, line 79: my problem with this is I don;t think that `b_partials` length `1<<(N+1)-1`
because it has length `N`
and those are never equal

oxinabox:
Nightly is failing just cos of reverse mode, unreleated to this PR

src/stage1/compiler_utils.jl
```diff
@@ -7,7 +7,7 @@ function Base.push!(cfg::CFG, bb::BasicBlock)
     push!(cfg.index, bb.stmts.start)
 end
 
-if VERSION <= v"1.11.0-DEV.116"
+if VERSION < v"1.11.0-DEV.258"
```
oxinabox, line 10: I am not sure where 116 came from but it wasn't working on the Cedar blessed just version which is greeater than 116.
SO I checked blame and found 
https://github.com/JuliaLang/julia/commit/517ad06a8e9f1f8ec3b80022fef5f32a717d6dd3
which is 258

src/stage1/recurse_fwd.jl
```diff
@@ -4,13 +4,51 @@ struct ∂vararg{N}; end
...
+            Tangent{iith_order_type, typeof(tangent_nt)}(tangent_nt)
+        end
+        return tangent
+    end
+    return TaylorBundle{N, B}(the_primal, the_partials)
+end
+
+_construct(::Type{B}, args) where B<:Tuple = B(args)
+# Hack for making things that do not have public constructors constructable:
+@generated _construct(B::Type, args) = :($(Expr(:splatnew, :B, :args)))
```
oxinabox, line 51:
```suggestion
@generated _construct(B::Type, args) = Expr(:splatnew, :B, :args)
```

Keno:
I don't exactly remember what the original motivation for this was, but if everything goes through fine, I have no particular attachment to this type.

src/stage1/compiler_utils.jl
oscardssmith: this came from me making up a number :).

src/stage1/forward.jl
```diff
@@ -70,26 +54,33 @@ function taylor_compatible(a::ATB{N}, b::ATB{N}) where {N}
...
+            end
+        end
+        return TaylorBundle{N+1}(the_primal, the_partials)
     else
-        return TangentBundle{N+1}(r.tup[1].primal,
-            (r.tup[1].tangent.partials..., primal(b),
-            ntuple(i->partial(b,i), 1<<(N+1)-1)...))
+        #XXX: am dubious of the correctness of this
+        a_partials = ntuple(i->partial(r, ii)[1], N)
+        b_partials = ntuple(i->partial(r, ii)[2], N)
```
oscardssmith, line 77: does this really not need `i`?

src/stage1/recurse_fwd.jl
```diff
@@ -4,13 +4,51 @@ struct ∂vararg{N}; end
...
+    return (∂☆new{N}())(B, a...)
 end
 
 struct ∂☆new{N}; end
 
-(::∂☆new{N})(B::Type, a::AbstractTangentBundle{N}...) where {N} =
-    CompositeBundle{N, B}(a)
+# we split out the 1st order derivative as a special case for performance
+# but the nth order case does also work for this
+function (::∂☆new{1})(B::Type, xs::AbstractTangentBundle{1}...)
```
oscardssmith, line 15: does the first order version of this actually give better performance?

test/forward.jl
oscardssmith: do we know what change broke it?

src/stage1/recurse_fwd.jl
oxinabox: Yes. I checked with benchmark tools.
It's not a huge jump, but it is a bit better.
And given this is the most common case

src/stage1/forward.jl
oxinabox, line 77:
```suggestion
        a_partials = ntuple(i->partial(r, i)[1], N)
        b_partials = ntuple(i->partial(r, i)[2], N)
```

test/tangent.jl
```diff
@@ -28,21 +28,22 @@ using Test
...
-    # hits `getindex(::CompositeBundle{Foo152}, ::TaylorTangentIndex)`
+    # https://github.com/JuliaDiff/Diffractor.jl/issues/152
+    # Though we have now removed the underlying cause, we keep this as a regression test just in case
     struct Foo152
         x::Float64
     end
 
     # Unit Test
-    cb = CompositeBundle{1, Foo152}((TaylorBundle{1, Float64}(23.5, (1.0,)),))
+    cb = TaylorBundle{1, Foo152}(Foo152(23.5), (Tangent{Foo152}(;x=1.0),))
```
staticfloat, line 38:
```suggestion
    cb = TaylorBundle{1, Foo152}(Foo152(23.5), (Tangent{Foo152}(;x=1.0),))
```

PR #217, opened 2023-09-21
WIP fix higher order. 3rd works

oxinabox:
This is the PR @oscardssmith  and I started working on to fix `recursive_sin'''(x)`
the core of it is that `shuffle_down` is wrong (AFAICT)
but also then other things got in the way.

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -11,10 +11,14 @@ primal(z::ZeroTangent) = ZeroTangent()
 
 first_partial(x) = partial(x, 1)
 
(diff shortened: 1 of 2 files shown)


src/stage1/forward.jl
```diff
@@ -150,14 +162,16 @@ function (::∂☆internal{N})(args::AbstractTangentBundle{N}...) where {N}
 end
 (::∂☆{N})(args::AbstractTangentBundle{N}...) where {N} = ∂☆internal{N}()(args...)
 
-
+ChainRulesCore.frule((_, ẋ))
```
oxinabox, line 165:
```suggestion
```

PR #219, opened 2023-09-28
Forward mode mutation support

### LINKED ISSUE #189: Support mutation for forward mode

oxinabox:
Closes #189 

Requires https://github.com/JuliaDiff/ChainRulesCore.jl/pull/626
and https://github.com/JuliaDiff/ChainRules.jl/pull/738
and https://github.com/JuliaDiff/ChainRules.jl/pull/739

### DIFF
--- a/src/codegen/forward_demand.jl
+++ b/src/codegen/forward_demand.jl
@@ -262,12 +262,12 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
             return transform!(ir, arg, order, maparg)
         elseif isa(arg, GlobalRef)
             @assert isconst(arg)
-            return ZeroBundle{order}(getfield(arg.mod, arg.name))
+            return zero_bundle{order}()(getfield(arg.mod, arg.name))
         elseif isa(arg, QuoteNode)
-            return ZeroBundle{order}(arg.value)
+            return zero_bundle{order}(){order}(arg.value)
         end
         @assert !isa(arg, Expr)
-        return ZeroBundle{order}(arg)
+        return zero_bundle{order}()(arg)
     end
 
     for (ssa, (order, custom)) in enumerate(ssa_orders)
@@ -307,7 +307,7 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
                     stmt = insert_node!(ir, ssa, NewInstruction(inst))
                 end
 
-                replace_call!(ir, SSAValue(ssa), Expr(:call, ZeroBundle{order}, stmt))
+                replace_call!(ir, SSAValue(ssa), Expr(:call, zero_bundle{order}(), stmt))
             elseif isa(stmt, SSAValue) || isa(stmt, QuoteNode)
                 inst[:inst] = maparg(stmt, SSAValue(ssa), order)
                 inst[:type] = Any
@@ -322,7 +322,7 @@ function forward_diff_no_inf!(ir::IRCode, to_diff::Vector{Pair{SSAValue,Int}};
                 inst[:type] = Any
                 inst[:flag] |= CC.IR_FLAG_REFINED
             else
-                val = ZeroBundle{order}(inst[:inst])
+                val = zero_bundle{order}()(inst[:inst])
                 inst[:inst] = val
                 inst[:type] = Const(val)
             end
--- a/src/stage1/recurse_fwd.jl
+++ b/src/stage1/recurse_fwd.jl
@@ -50,7 +50,7 @@ _construct(::Type{B}, args) where B<:Tuple = B(args)
 # Hack for making things that do not have public constructors constructable:
 @generated _construct(B::Type, args) = Expr(:splatnew, :B, :args)
 
-@generated (::∂☆new{N})(B::Type) where {N} = return :(ZeroBundle{$N}($(Expr(:new, :B))))
+@generated (::∂☆new{N})(B::Type) where {N} = return :(zero_bundle{order}()($(Expr(:new, :B))))
 
 # Sometimes we don't know whether or not we need to the ZeroBundle when doing
 # the transform, so this can happen - allow it for now.
--- a/src/tangent.jl
+++ b/src/tangent.jl
@@ -378,3 +378,18 @@ end
 function ChainRulesCore.rrule(::typeof(rebundle), atb)
     rebundle(atb), Δ->throw(Δ)
 end
+
+
+"""
+    (::zero_bundle{N})(primal)
+
+Creates a bundle with a zero tangent.
+"""
+struct zero_bundle{N} end
+function (::zero_bundle{N})(primal) where N
+    if zero_tangent(primal) isa ZeroTangent
+        return ZeroBundle{N}(primal)
+    else
+        return TaylorBundle{N}(primal, ntuple(_->zero_tangent(primal), N))
+    end
+end
--- a/test/forward_mutation.jl
+++ b/test/forward_mutation.jl
@@ -1,0 +1,16 @@
+using Diffractor
+using Diffractor: ∂☆
+using Diffractor: bundle
+
+mutable struct MDemo1
+    x::Float64
+end
+function double!(val::MDemo1)
+    val.x *= 2.0
+    return val
+end
+function wrap_and_double(x)
+    val = MDemo1(x)
+    double!(val)
+end
+∂☆{1}()(ZeroBundle{1}(wrap_and_double), TaylorBundle{1}(1.5, (1.0,)))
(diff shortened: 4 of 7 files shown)


test/forward_mutation.jl
```diff
@@ -0,0 +1,51 @@
...
+@test 🐇[TaylorTangentIndex(2)] isa MutableTangent
+@test 🐇[TaylorTangentIndex(2)] == 0.0  # returns 20
+
+
+
+foo(val) = val^2
+🥖 = ∂☆{2}()(ZeroBundle{2}(foo), TaylorBundle{2}(1.0, (0.0, 10.0)))
+🥖[TaylorTangentIndex(1)] # returns 0
+🥖[TaylorTangentIndex(2)] # returns 20
+
```
oxinabox, line 50:
```suggestion
```

test/forward_mutation.jl
```diff
@@ -0,0 +1,51 @@
...
+🐰 = ∂☆{1}()(ZeroBundle{1}(wrap_and_double), TaylorBundle{1}(1.5, (1.0,)))
+@test first_partial(🐰) isa MutableTangent{MDemo1}
+@test first_partial(🐰).x == 2.0
+
+# second derivative
+🐇 = ∂☆{2}()(ZeroBundle{2}(wrap_and_double), TaylorBundle{2}(1.5, (1.0, 10.0)))
+@test 🐇[TaylorTangentIndex(1)] isa MutableTangent{MDemo1}
+@test 🐇[TaylorTangentIndex(1)].x == 2.0
+@test 🐇[TaylorTangentIndex(2)] isa MutableTangent
+@test 🐇[TaylorTangentIndex(2)] == 0.0  # returns 20
```
oxinabox, line 42: @Keno  any idea why this is happening?
Do I need to do some un-taylor step?
Like 
```julia
deriviative_2(🐇::TaylorBundle{2}) = 🐇[TaylorTangentIndex(2)]/2 - 🐇[TaylorTangentIndex(2)])
```
(assuming seeded with 1.0,1.0, which this isn't)

If so is there a function in here for applying that?

oxinabox:
Changes in the ChainRulesCore PR make `zero_tangent` more aggressive at making structural zeros.
or `NoTangent`s for things without tangent spaces (in particular if they don't have any fields)
This is good because it should fix up a bunch of our type inference issues, since were already using `zero_tangent` everywhere in this PR.
A complicating factor is we also need to fix up where we use `ZeroTangent` for thigns that actually have no tangent space -- which isn't wrong as such but if we don't care either way we should be disaptching on `AbstractZero`.
Which this PR now does.

oxinabox:
I merged #236 in here because i was running into code that was hitting the exact case that fixes in the tests.
Which does show our code path is now significantly different, probably because of the structural zeros.

src/codegen/forward_demand.jl
```diff
@@ -105,6 +105,7 @@ function forward_diff_uncached!(ir::IRCode, interp::AbstractInterpreter, irsv::I
             Δbacking = insert_node!(ir, ssa, NewInstruction(Expr(:splatnew, widenconst(tup_typ), Δbacking), tup_typ_typ.val))
         end
         tangentT = Core.Compiler.apply_type_tfunc(Const(ChainRulesCore.Tangent), newT, tup_typ_typ).val
+        # TODO do we need to make sure this inserts right
```
oxinabox, line 108: What does this mean?

src/codegen/forward_demand.jl
```diff
@@ -362,6 +363,6 @@ function forward_diff!(interp::ADInterpreter, ir::IRCode, src::CodeInfo, mi::Met
     rt = CC._ir_abstract_constant_propagation(interp, irsv)
 
     ir = compact!(ir)
-
+
```
oxinabox, line 366:
```suggestion
```

src/stage1/recurse_fwd.jl
```diff
@@ -50,7 +50,7 @@ _construct(::Type{B}, args) where B<:Tuple = B(args)
 # Hack for making things that do not have public constructors constructable:
 @generated _construct(B::Type, args) = Expr(:splatnew, :B, :args)
 
-@generated (::∂☆new{N})(B::Type) where {N} = return :(ZeroBundle{$N}($(Expr(:new, :B))))
+@generated (::∂☆new{N})(B::Type) where {N} = return :(zero_bundle{order}()($(Expr(:new, :B))))
```
oxinabox, line 53:
```suggestion
@generated (::∂☆new{N})(B::Type) where {N} = return :(zero_bundle{$N}()($(Expr(:new, :B))))
```

src/stage1/recurse_fwd.jl
```diff
@@ -15,31 +15,31 @@ struct ∂☆new{N}; end
...
-        
     the_partials = ntuple(Val{N}()) do ii
-        iith_order_type = ii==1 ? B : Any  # the type of the higher order tangents isn't worth tracking
         tangent_tup = map(x->partial(x, ii), xs)
         tangent = if B<:Tuple
-            Tangent{iith_order_type, typeof(tangent_tup)}(tangent_tup)
+            Tangent{B, typeof(tangent_tup)}(tangent_tup)
         else
+            # It is a little dubious using StructuralTangent{B} for >1st order, but it is isomorphic.
+            # Just watch out for order mixing bugs.
```
oxinabox, line 39: I am not even sure it is actually dubious. Further thought has convinced me its the right thing to do.
Ironically (?) at higher order it is not a structural tangent but rather a Natural tangent.
But eitherway it follows the requirement that it can be added to its base space to get a nother value in the base space. Its like `Real` numbers in that way,

src/codegen/forward_demand.jl
oscardssmith: I think we wanted to make sure that it was inserting the node in the right spot in the IR (specifically in the presence of control flow)

test/reverse.jl
```diff
@@ -98,14 +98,16 @@ let var"'" = Diffractor.PrimeDerivativeBack
...
+    # Control flow cases:
+    # if
     @test @inferred((x->simple_control_flow(true, x))'(1.0)) == sin'(1.0)
     @test @inferred((x->simple_control_flow(false, x))'(1.0)) == cos'(1.0)
     @test (x->sum(isa_control_flow(Matrix{Float64}, x)))'(Float32[1 2;]) == [1.0 1.0;]
-    @test times_three_while'(1.0) == 3.0
-
+    
+    # while
+    # @test times_three_while'(1.0) == 3.0  # hangs in 1.11
```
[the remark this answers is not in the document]
oxinabox: No, but its unrelated to this PR

src/codegen/forward_demand.jl
oxinabox: Ok seems unreleated to this PR

```suggestion
```

src/extra_rules.jl
```diff
@@ -172,9 +172,14 @@ function ChainRules.rrule(::DiffractorRuleConfig, ::Type{SArray{S, T, N, L}}, x:
 end
 
 function ChainRules.frule((_, ∂x), ::Type{SArray{S, T, N, L}}, x::NTuple{L,T}) where {S, T, N, L}
-    SArray{S, T, N, L}(x), SArray{S, T, N, L}(∂x.backing)
+    #TODO: we really shouldn't actually see the isa(∂x, AbstractZero) case since the frule should be called then
+    Δx = isa(∂x, AbstractZero) ? ∂x : SArray{S, T, N, L}(ChainRulesCore.backing(∂x))
```
oxinabox, line 176: this isn't hit anymore
```suggestion
    Δx = SArray{S, T, N, L}(ChainRulesCore.backing(∂x))
```

src/stage1/recurse_fwd.jl
oxinabox, line 39:
```suggestion
            # No matter the order we use `StructuralTangent{B}` for the partial
            # It follows all required properties of the tangent to the n-1th order tangent
```

src/tangent.jl
```diff
@@ -418,3 +418,20 @@ end
...
+
+
+"""
+    (::zero_bundle{N})(primal)
+
+Creates a bundle with a zero tangent.
+"""
+struct zero_bundle{N} end
+function (::zero_bundle{N})(primal) where N
+    if zero_tangent(primal)  isa AbstractZero
```
oxinabox, line 430:
```suggestion
    # We still use a Uniform bundle e.g. if primal has NoTangent
    if zero_tangent(primal) isa AbstractZero
```

PR #221, opened 2023-10-04
Signal Measurement Demo

oxinabox:
This is I think a good demo of a key capability.
Making this work should make sure a variety of more complex variations work.

I am not sure if this itself will work at all without a custom frule to get around the fact that we are working with sampled data rather than a smooth function.
So this is here to test that.

However, I am currently running into other issues, it seems we can't actually AD broadcast.

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -202,6 +202,7 @@ function (::∂☆{N})(::ZeroBundle{N, typeof(map)}, f::ATB{N}, args::ATB{N, <:A
     rebundle(map(FwdMap(f), map(unbundle, args)...))
 end
 
+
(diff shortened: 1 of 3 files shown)


src/stage1/forward.jl
```diff
@@ -203,6 +203,7 @@ function (::∂☆{N})(::ZeroBundle{N, typeof(map)}, f::ATB{N}, args::ATB{N, <:A
     rebundle(map(FwdMap(f), map(unbundle, args)...))
 end
 
+
```
oxinabox, line 206:
```suggestion
```

oxinabox:
Having implemented this downstream in a very different way (using Implict function theorem, rather than on sampled space) this PR is no longer useful to me

Issue #222, opened 2023-10-06
Unable to pass config to call back into AD


rcalxrc08:
I have the following:
```julia
function my_func(x)
	return sin(x)
end
using ChainRulesCore
using AbstractDifferentiation: derivative, jacobian
using Diffractor: DiffractorForwardBackend, DiffractorRuleConfig
function ChainRulesCore.frule(config::RuleConfig{>:HasForwardsMode}, (_, dx), ::typeof(my_func), x)
    res=my_func(x)
    _, der_fwd = ChainRulesCore.frule_via_ad(config, (ChainRulesCore.NoTangent(), dx), sin, x)
    @show "hello from fd with right config with calling back"
    return res, der_fwd
end

function my_func2(x)
	return sin(x)
end

function ChainRulesCore.frule((_, dx), ::typeof(my_func2), x)
    res=my_func2(x)
    _, der_fwd = ChainRulesCore.frule_via_ad(DiffractorRuleConfig(), (ChainRulesCore.NoTangent(), dx), sin, x)
    @show "hello from fd with wrong config with calling back"
    return res, der_fwd
end

derivative(DiffractorForwardBackend(), my_func, 1.0)
derivative(DiffractorForwardBackend(), my_func2, 1.0)
```
and I get the output just for the second call, in the first case the frule has something wrong. Is it supported this way of defining the rules for Diffractor.jl? Am I doing anything wrong?

oxinabox:
You are not, you are doing everything correct.
Its just that right now diffractor forward mode isn't using its config.
I am not sure why it was commented out.
See here: https://github.com/JuliaDiff/Diffractor.jl/blob/b3e4ee058ef0ae59bc5c7623ed6770b52b6324f2/src/stage1/forward.jl#L110C17-L110C37
Could be we can just uncomment it out and add a test and be good to go.
It might have just been commented out while investigating something and forgotten to undone.
Or it might be something more annoying like inference issues.

PR #224, opened 2023-10-10
Fix map in forwards mode

oxinabox:
This is sliced out of #221 
since that stuff is going to take a bit more thinking.

but this is an obvious and easy bug fix.
Both tests errored before these changes and now they do not.

Arguably we should delete Diffractors sepecial handling of `map` all together and leave that to ChainRules.
But that is a bigger PR.
Also requires fixing #222  (which might be easy)

### DIFF
--- a/src/tangent.jl
+++ b/src/tangent.jl
@@ -206,20 +206,25 @@ end
 
 const TaylorBundle{N, B, P} = TangentBundle{N, B, TaylorTangent{P}}
 
+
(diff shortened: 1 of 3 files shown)


oscardssmith:
this looks good, but you should probably define equality and hashing such that different representations of the same bundle end up equal

oxinabox:
the PR to do that ended up being more involved than you would think.
So I will make it as a follow up

PR #225, opened 2023-10-11
Add polish to Bundle Types

oxinabox:
This PR does several things that make basically no user facing change.
It was the follow up to #224 but actually ended up way more involved.

 - Adds the missing `ExplictTangentBundle[CanonicalTangentIndex(i)]`  (which is funny that it was missing as that's the natural way to index an `ExplictTangentBundle`
 - Moves the `getindex` main definitions down on to to the tangent types.
 - Defines promotion rules: **Explict > Taylor > Uniform**
 - Defines conversions. For now only `Uniform->Taylor/Explict` and `Taylor->Explict`, though in future we could put the other way around in (but they are not always going to be defined)
 - Defines equality for everything, via defining self equality and then a fallback to promotion  (apparently this is the way the julia docs recommend defining equality. I have always just defined it directly)
 - Defines `hash` for everything, this just uses the `hash` of the primal for bundles, which I think is actually sufficient.  I am making the assumption that we do not end up needing a dict which stores the same primal with different tangents a ton of times. We can revisit that later if it comes up.
 - Deletes `ProductTangent`, as we were not using it, and I didn't want to implement the above for it. (I think `ProductTangent` was basically the same as `CompositeTangent` but less complete, and that's gone now anyway)

### DIFF
--- a/src/tangent.jl
+++ b/src/tangent.jl
@@ -74,11 +74,8 @@ struct TaylorTangentIndex <: TangentIndex
     i::Int
 end
 
-function Base.getindex(a::AbstractTangentBundle, b::TaylorTangentIndex)
-    error("$(typeof(a)) is not taylor-like. Taylor indexing is ambiguous")
-end
-
 abstract type AbstractTangentSpace; end
+Base.:(==)(x::AbstractTangentSpace, y::AbstractTangentSpace) = ==(promote(x, y)...)
 
 """
     struct ExplicitTangent{P}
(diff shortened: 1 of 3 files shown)


src/tangent.jl
```diff
@@ -89,13 +86,23 @@ represented by a vector of `2^N-1` partials.
...
 end
+Base.:(==)(a::ExplicitTangent, b::ExplicitTangent) = a.partials == b.partials
+Base.hash(tt::ExplicitTangent, h::UInt64) = hash(tt.partials, h)
+
+Base.getindex(tangent::ExplicitTangent, b::CanonicalTangentIndex) = tangent.partials[b.i]
+function Base.getindex(tangent::ExplicitTangent, b::TaylorTangentIndex)
+    if lastindex(tangent.partials) == exp2(b.i) - 1
+        return tangent.partials[end]
+    end
+    # TODO: should we also allow other indexes if all the partials at that level are equal up regardless of order?
```
oxinabox, line 97: the comment wasn't in there before.
But I don't actually see why we don't do this.

We just need to check all things at the indexes `j` that have `count_ones(j) = b.i` are equal. And if so return that thing.
if not give error about ambig

oscardssmith (approved):

PR #229, opened 2023-10-13
make `shuffle_down` of `TaylorBundle` use `TaylorBundle` rather than `ExplicitBundle`

oscardssmith:
@Keno does this look right to you?


oxinabox (approved):
This is fine. `TaylorBundle{1}` and `ExplictBundle{1}` are the same type and the `TaylorBundle` case is much more well tested.

PR #234, opened 2023-10-17
Update to changes to construct_ssa

oxinabox:
This is an update to make things work after https://github.com/JuliaLang/julia/pull/50943

Right now however it seems to makes tests hang

### DIFF
--- a/src/stage1/recurse.jl
+++ b/src/stage1/recurse.jl
@@ -276,7 +276,13 @@ function optic_transform!(ci, mi, nargs, N)
     stmts = VERSION < v"1.11.0-DEV.258" ? ir.stmts.inst : ir.stmts.stmt
     defuse_insts = scan_slot_def_use(Int(meth.nargs), ci, stmts)
     ci.ssavaluetypes = Any[Any for i = 1:ci.ssavaluetypes]
-    ir = construct_ssa!(ci, ir, domtree, defuse_insts, ci.slottypes, SimpleInferenceLattice.instance)
+    if VERSION > v"1.11.0-DEV.337"
+        interp = NativeInterpreter(meth.primary_world)
+        opt_state=OptimizationState(mi, ci, interp)
+        ir = construct_ssa!(ci, ir, opt_state, domtree, defuse_insts, SimpleInferenceLattice.instance)
+    else
+        ir = construct_ssa!(ci, ir, domtree, defuse_insts, ci.slottypes, SimpleInferenceLattice.instance)
+    end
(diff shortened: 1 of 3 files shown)


src/stage1/recurse.jl
```diff
@@ -276,7 +276,13 @@ function optic_transform!(ci, mi, nargs, N)
     stmts = VERSION < v"1.11.0-DEV.258" ? ir.stmts.inst : ir.stmts.stmt
```
topolarity, line 276: FWIW in the typical pipeline `convert_to_ircode` is used to convert CodeInfo-style statements to IRCode, but that's skipped here.

Most of the function is a no-op if you have no (informative) results from inference but there is some normalization (e.g. `strip_trailing_junk!`) that might be needed here for certain IR.

None of that is new in this PR though, so we can worry about it later.

oxinabox:
@topolarity any ideas why this would make it hang?
Seen both locally and in CI

topolarity:
Not sure, but here's a reduced MWE at least:
```julia
using Diffractor

function times_three_while(x)
    z = x
    i = 3
    while i > 1
        z += x
        i -= 1
    end
    z
end

let var"'" = Diffractor.PrimeDerivativeBack
    println(times_three_while'(1.0) == 3.0)
end
```

The IR that Diffractor generates is identical on latest Julia versus `2cee483bce` ([ir.txt](https://github.com/JuliaDiff/Diffractor.jl/files/13025593/ir.txt)), but inference appears to get stuck in a loop doing semi-concrete eval of `>` and `-`:
```
semi-concrete eval: (::Diffractor.∂⃖{1})(typeof(Base.:(-)), Int64, Int64) from (::Diffractor.∂⃖{N})(T, Any...) where {T, N}
semi-concrete eval: (::Diffractor.∂⃖{1})(typeof(Base.:(>)), Int64, Int64) from (::Diffractor.∂⃖{N})(T, Any...) where {T, N}
semi-concrete eval: (::Diffractor.∂⃖{1})(typeof(Base.:(-)), Int64, Int64) from (::Diffractor.∂⃖{N})(T, Any...) where {T, N}
semi-concrete eval: (::Diffractor.∂⃖{1})(typeof(Base.:(>)), Int64, Int64) from (::Diffractor.∂⃖{N})(T, Any...) where {T, N}
... (repeats infinitely)
```

oxinabox:
I guess since the IR is identical, we could consider merging this.

aviatesk:
Going to merge as the 1.10 CI passes.

PR #235, opened 2023-10-17
handle Expr(:isdefined,) in forwards mode

oxinabox:
this was found when doing a nasty nested AD thing.
I do think its nice that https://github.com/JuliaDiff/Diffractor.jl/pull/167 is paying off and making these easy to find and fix


oxinabox:
@oscardssmith  bumping for review

oscardssmith (approved):

oscardssmith:
I also realize this is somewhat unrelated, but why is `isdefined` a head, rather than just being a `:call`?

oxinabox:
I assume because we sometimes reason about it in the optimizer.

PR #236, opened 2023-10-17
Handle tuples with types in them

oxinabox:
Diffractor was having a bit of a sad because `(Int,)` has type `Tuple{DataType}`
but the primal `B`  has type `Tuple{Type{Int}}`,
and while they are logically the same apparently dispatch does not like that.

The solution right now in this PR is to throw away that extra info we have about `B` and normalize it to just `Typle{DataType}`.
idk if it is best, I am open to other ideas.


oxinabox:
@Keno could you give this a review, and see if you think of a better way?

staticfloat:
@Keno review please

PR #239, opened 2023-10-24
renable configured frules

### LINKED ISSUE #222: Unable to pass config to call back into AD

oxinabox:
Resolves #222 
I did a quick test and this didn't seem to break inference.
So I think this is fine.
I don't see reasons it shouldn't work.
But I also do not know why it was disabled in the first place.

at some point in future I think will want to AD through root-finding.
and Roots.jl uses a configured `frule` for that.

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -106,7 +106,7 @@ function shuffle_base(r)
 end
 
 function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
-    r = frule(#=DiffractorRuleConfig(),=# map(first_partial, args), map(primal, args)...)
+    r = frule(DiffractorRuleConfig(), map(first_partial, args), map(primal, args)...)
     if r === nothing
         return ∂☆recurse{1}()(args...)
     else
--- a/test/forward.jl
+++ b/test/forward.jl
@@ -184,4 +184,22 @@ end
     )
 end
 
+
+@testset "configured frule" begin
+    my_func(x) = sin(x)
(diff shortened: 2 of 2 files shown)


oxinabox:
Keno said yes in a call, so am merging this

PR #254, opened 2024-01-29
Start testing reverse mode

nmheim:
Adds tests for `Base` taken from Zygote. There will be quite a few more tests coming in, probably in a new PR (essentially planning to updated/split #73 into smaller MRs). Many of the added tests are currently broken (with the error message in the comment above the broken test).

Two main takeaways so far are: kwargs seem to break tests with one of two errors:
```
    # MethodError: no method matching copy(::Nothing)
    # Rewrite reached intrinsic function bitcast. Missing rule?
```
and then there are a few tests that return incorrect gradients.

Is there a good place to keep the main problems of the broken tests?

In #73 the `pullback` function from Zygote is kept, should we keep it or rewrite those tests in a different way?


oxinabox:
> In https://github.com/JuliaDiff/Diffractor.jl/pull/73 the pullback function from Zygote is kept, should we keep it or rewrite those tests in a different way?

Let's rename it to `value_and_pullback` following the name it will get once we hook it up to AbstractDifferentiation.

FYI: he name "hobbit" for this is a reference to to
> The Hobbit, or There and Back Again is a children's fantasy novel by English author J. R. R. Tolkien.

Hobbit == There and Back Again

oxinabox:
>  Is there a good place to keep the main problems of the broken tests?

I tend to do what you are doing
Comment above.
Plus I also will normally create an issue on the repo and link to it in the comment if we are going to merge with the tests still marked as broken

src/stage1/generated.jl
```diff
@@ -320,7 +320,7 @@ function (::∂⃖{N})(::typeof(Base.getindex), a::Array{<:Number}, inds...) whe
             (@Base.constprop :aggressive Δ->begin
                 Δ isa AbstractZero && return (NoTangent(), Δ, map(Returns(Δ), inds)...)
                 BB = zero(a)
-                BB[inds...] = unthunk(Δ)
+                view(BB, inds...) .+= unthunk(Δ)
```
oxinabox, line 323: why is this different?
`0 .+ Δ = Δ` at least assuming they are the same size

oxinabox:
I think we should merge this more or less as is and then can continue with follow up PRs to add more tests / fix existing broken ones.
What do you think?

src/stage1/generated.jl
nmheim: Ah sorry, I should have added a comment here. this is for cases like
```julia
julia> gradient(sum ∘ x -> x[:,[1,1,2]], rand(3,4))[1]
3×4 Matrix{Float64}:
 2.0  1.0  0.0  0.0
 2.0  1.0  0.0  0.0
 2.0  1.0  0.0  0.0
```
The `view` lets one add duplicate indices
```julia
julia> x = zeros(3,4)
3×4 Matrix{Float64}:
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0

julia> view(x, :, [1,1,2]) .+= ones(3,3)
3×3 view(::Matrix{Float64}, :, [1, 1, 2]) with eltype Float64:
 2.0  2.0  1.0
 2.0  2.0  1.0
 2.0  2.0  1.0

julia> x
3×4 Matrix{Float64}:
 2.0  1.0  0.0  0.0
 2.0  1.0  0.0  0.0
 2.0  1.0  0.0  0.0
```

nmheim:
> I think we should merge this more or less as is and then can continue with follow up PRs to add more tests / fix existing broken ones. What do you think?

Yes, that sounds good! I can also open another PR with the `view` fix. would it be better to have a non-mutating version of it?

src/stage1/generated.jl
oxinabox: oh cool trick. I think last time i needed to handle that my solution was way more complicated.

oxinabox:
> maybe it would be better to have a non-mutating version of it?

It's fine for now.
Can do that later

oxinabox (approved):

PR #255, opened 2024-01-31
Update to new AbstractDifferentiation.jl

MasonProtter:
Potential alternative to https://github.com/JuliaDiff/Diffractor.jl/pull/228

There's something weird and wrong happening with failing tests though. If I include all the stuff into my REPL, `@test_broken` behaves properly, but if I actually do `] test` then a bunch of them propagate their failures outside of the `@test_broken` so I had to manually comment them out to get them to behave.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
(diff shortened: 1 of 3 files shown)


MasonProtter:
bump @oxinabox

PR #257, opened 2024-02-01
Allow CI failure on nightly

gdalle:
I guess with Diffractor tests on nightly are +/- doomed to fail, so they shouldn't get in the way of automated PR checks passing and the badge being green

See https://discourse.julialang.org/t/ignoring-nightly-failure-for-ci-badge/98028/5?u=gdalle


gdalle:
I don't see what went wrong here, just copy-pasted from Graphs.jl where it worked like a charm

https://github.com/JuliaGraphs/Graphs.jl/blob/master/.github/workflows/ci.yml

gdalle:
I think the badge is still red because the documentation fails https://github.com/JuliaDiff/Diffractor.jl/actions/runs/7742171459/job/21110758210

PR #263, opened 2024-02-06
Eras mode

### LINKED ISSUE #252: Eras mode: Taylor or it Eras
Right now if a particular derivative can not be represented as a TaylorBundle we fall back and represent it as a ExplictTangentBundle.

For the last 6 months we have been trying the other option: if it can't be represented as a TaylorBundle erroring out in various branches that include `for_cedar` in the name.

This is kinda annoying since it will error sometimes -- but you can generally fix that (if you are running your own optimization passes) by optimizing until all the diffractor code vanishes, before doing the next level of nested AD.
But it has the key advantage of being much faster since it avoids the type instability between TaylorTangent vs ExplictTangents

Anyway this maintainign this feature in a seperate branch is really annoying.
So we should add a "Eras mode" type parameter to ` ∂☆` to tell if if we should error or use ExplictTangents.

To be safe in Eras mode,  
it is sufficient, but not necessary, to be doing nested AD with respect to the same variable -- this naturally transforms into a Taylor AD.
It also works in other cases where (likely by problem construction) ADing with respect to a second variable happens to result in something that can be represented with a TaylorBundle also. (You need your different partials to happen to be exactly equal).

oxinabox:
Closes #252 

- [x] changes to the internals
- [x] hooking up for forward_demand
- [x] tests

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -257,7 +257,7 @@ function (::∂☆{N})(f::ATB{N, typeof(tuple)}, args::AbstractTangentBundle{N}.
     ∂vararg{N}()(args...)
 end
 
-function (::∂☆{N})(f::ATB{N, typeof(tuple)}, args::ZeroBundle{N}...) where {N}
+function (::∂☆{N})(f::ATB{N, typeof(tuple)}, args::AbstractZeroBundle{N}...) where {N}
     ZeroBundle{N}(map(primal, args))  # special fast case
 end
 
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -95,13 +95,14 @@ function shuffle_up(r::UniformBundle{N, B, U}) where {N, B, U}
 end
 @ChainRulesCore.non_differentiable shuffle_up(r::UniformBundle)
 
-function shuffle_up_bundle(r::Diffractor.TangentBundle{1, <:TaylorBundle{1}})
+
+function shuffle_up_bundle(r::Diffractor.TangentBundle{1, B}) where {B<:ATB{1}}
     a = primal(r)
     b = partial(r, 1)
     z₀ = primal(a)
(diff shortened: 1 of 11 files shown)


oxinabox:
I have started adding tests but right now they show no change unnested,
and once we nest type inference is failing with or without eras mode being turned on.
So that needs to be fixed.
I suspect it is because I am using a if statement rather than a seperate method, in `shuffle_up`

src/stage1/forward.jl
```diff
@@ -108,32 +199,37 @@ function shuffle_base(r)
...
+        # If all partials are immutable AbstractZero subtyoes we know we don't have to worry about a mutating frule either
+        r = f(primal_args...)
+        return r, zero_tangent(r)
+    end
 end
 
 function ChainRulesCore.frule_via_ad(::DiffractorRuleConfig, partials, args...)
-    bundles = map(bundle, args, partials)
-    result = ∂☆internal{1}()(bundles...)
+    bundles = map(bundle, partials, args)
```
oxinabox, line 224:
```suggestion
    bundles = map(bundle, args, partials)
```

oxinabox:
A question for the reviewer.
Is this a breaking change?
It doesn't break the public API. and technically doesn't break calling the private `∂☆{1}()(...)` API
So I guess not?

oxinabox:
idk why Diffractor isn't working one 1.11

oxinabox:
- 1.11-nightly is broken due to https://github.com/JuliaDebug/Cthulhu.jl/issues/541 which is fixed on release-1.11, so we are waiting for a new julia release to be cut there
 - nightly is broken for reasons related to this PR I don't yet undestand. I can't reproduce with my (slightly outdated) version of nightly.

oxinabox:
fixed everything, this is now just waiting for review

Keno (approved):
Seems good to me.

PR #266, opened 2024-02-09
Support frules with keyword arguments

### LINKED ISSUE #244: support for frules with keyword arguments
I wrote an frule with keyword arguments
It doesn't look like diffractor is picking it up.

oxinabox:
Closes #244

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -117,11 +117,33 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
     end
 end
 
-_frule(partials, primals...) = frule(DiffractorRuleConfig(), partials, primals...)
-function _frule(::NTuple{<:Any, AbstractZero}, f, primal_args...)
+# frules with keywords support:
+function (::∂☆internal{1})(kwc::ATB{1, typeof(Core.kwcall)}, kw::ATB{1}, f::ATB{1}, args::ATB{1}...)
+    args_primals = (primal(f), map(primal, args)...)
+    args_partials = (first_partial(f), map(first_partial, args)...)
+    # First check if directly overloading frule for kwcall
+    r = _frule(
+        (first_partial(kwc), first_partial(kw), args_partials...),
+        primal(kwc), primal(kw), args_primals...
+    )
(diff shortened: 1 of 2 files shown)


src/stage1/forward.jl
```diff
@@ -117,11 +117,33 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
...
+    # First check if directly overloading frule for kwcall
+    r = _frule(
+        (first_partial(kwc), first_partial(kw), args_partials...),
+        primal(kwc), primal(kw), args_primals...
+    )
+    if r===nothing
+        # then check if the frule for f accepts keywords
+        # This silently discards tangents of the kw-args
+        # TODO: should we error if they nonzero?
+        r = _frule(args_partials, args_primals...; primal(kw)...)
```
Keno, line 133: I'm not sure I like this fallback. The problem is that if we support it at first order, we need to have the same logic at higher orders and it just becomes complicated.

src/stage1/forward.jl
```diff
@@ -117,11 +117,33 @@ function (::∂☆internal{1})(args::AbstractTangentBundle{1}...)
...
+    end
+    if r === nothing
+        return ∂☆recurse{1}()(kwc, kw, f, args...)
+    else
+        return shuffle_base(r)
+    end
+end
+
+_frule(partials, primals...; kwargs...) = frule(DiffractorRuleConfig(), partials, primals...; kwargs...)
+function _frule(::NTuple{<:Any, AbstractZero}, f, primal_args...; kwargs...)
```
Keno, line 143: I suspect for performance, you probably want to define the non-kwargs version separately, so that higher-order AD doesn't have to AD through all the kwargs dispatch logic, like we do in ChainRules.

src/stage1/forward.jl
oxinabox: If we don't have this, then we do not need this PR at all, since writing the rule for `kwcall` currently does work.
This particular logic is the code of this PR.
It's what makes it possible to write frules that have keywords args that match to functions with keyword args.


I don't understand what the problem with this at higher order is?

PR #268, opened 2024-02-12
Adjust to Cthulhu API change

Keno:


oxinabox:
This change was made in https://github.com/JuliaDebug/Cthulhu.jl/pull/522
`v"2.10.1"` was the first release after that, do we need to add such a lower bound on Cthulhu ?

oxinabox (approved):

aviatesk (approved):
:+1 for @oxinabox 's suggestion.

PR #270, opened 2024-02-13
Use rrule of KwFunc for Core.kwcall

nmheim:
This seems to fix non-differentiable keyword arguments.

### DIFF
--- a/src/stage1/generated.jl
+++ b/src/stage1/generated.jl
@@ -206,20 +206,20 @@ function (::∂⃖{N})(f::Core.IntrinsicFunction, args...) where {N}
 end
 
 # The static parameter on `f` disables the compileable_sig heuristic
-function (::∂⃖{N})(f::T, args...) where {T, N}
+function (::∂⃖{N})(f::T, args...; kwargs...) where {T, N}
     if N == 1
         # Base case (inlined to avoid ambiguities with manually specified
         # higher order rules)
-        z = rrule(DiffractorRuleConfig(), f, args...)
+        z = rrule(DiffractorRuleConfig(), f, args...; kwargs...)
         if z === nothing
-            return ∂⃖recurse{1}()(f, args...)
+            return ∂⃖recurse{1}()(f, args...; kwargs...)
         end
         return z
     else
         ∂⃖p = ∂⃖{N-1}()
-        @destruct z, z̄ = ∂⃖p(rrule, f, args...)
+        @destruct z, z̄ = ∂⃖p(rrule, f, args...; kwargs...)
         if z === nothing
-            return ∂⃖recurse{N}()(f, args...)
+            return ∂⃖recurse{N}()(f, args...; kwargs...)
         else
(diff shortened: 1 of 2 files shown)


src/stage1/generated.jl
```diff
@@ -244,6 +244,10 @@ struct KwFunc{T,S}
 end
 (kw::KwFunc)(args...) = kw.kwf(args...)
 
+function ChainRulesCore.rrule(::typeof(Core.kwcall), kwargs, f, args...)
+    rrule(KwFunc(f), kwargs, f, args...)
```
oxinabox, line 248: why isn't this
```suggestion
    rrule(f, args...; kwargs...)
```

oxinabox:
@Keno  I think this does *not* have the same concerns that #266 has, but I am not sure.

src/stage1/generated.jl
nmheim: Because (I think) we want to hit this `rrule` for `KwFunc`:
https://github.com/JuliaDiff/Diffractor.jl/pull/270/files#diff-5bd76352c0319c1d2659edc9825a90b92ac204aa1c17b9b95a3880e0fd98855eR255

nmheim: I am not exactly sure why the `KwFunc` struct is needed though... maybe it could be done directly via `rrule(::typeof(Core.kwcall), ...)` directly?

nmheim: Removing the `KwFunc` seems and dispatching on `kwcall` directly seems to work, but I was afraid to remove something which I don't exactly understand
```julia
function ChainRulesCore.rrule(::typeof(Core.kwcall), kwargs, f, args...)
    r = Core.kwfunc(rrule)(kwargs, rrule, f, args...)
    if r === nothing
        return nothing
    end
    x, back = r
    x, Δ->begin
        (NoTangent(), NoTangent(), back(Δ)...)
    end
end
```

oxinabox: Oh this might be the thing that is there to avoid ADing through so much of the kwarg machinery

Keno:
Same as in #266 I think in order for this not to make inference worse, the method should be split into kw and non-kw versions.

src/stage1/generated.jl
```diff
@@ -225,6 +225,25 @@ function (::∂⃖{N})(f::T, args...) where {T, N}
         end
     end
 end
+function (::∂⃖{N})(::typeof(Core.kwcall), kwargs, f::T, args...) where {T, N}
+    if N == 1
+        # Base case (inlined to avoid ambiguities with manually specified
+        # higher order rules)
+        z = rrule(DiffractorRuleConfig(), KwFunc(f), kwargs, f, args...)
```
nmheim, line 232: its basically a copy of the non-kw version of the function, but that is what we want in order to avoid ADing through the kw machinery, if there are no kws if I understand correctly?

PR #274, opened 2024-02-27
add ne_float intrinsic

pepijndevos:


oxinabox:
This PR is AFAICT correct.
I do not understand why we encounter this however.
Still I do not see the harm in having it

oxinabox:
I have also cherry-picked this on to the `ox/for_cedar/mutation2` branch.
so you should see it if you do `] up Diffractor`  on any of the projects using that branch also

PR #275, opened 2024-03-05
Fix frule for static array constructor that converts eltype

oxinabox:
Previously for the convertable case this errored with
```
  DimensionMismatch: No precise constructor for SVector{3, Float64} found. Length of input was 1
```

This does make use of the general assumption of ChainRulesCore that Number's are their own tangent type.
But I think that is fine, that assumption exists in a whole bunch of places and if you violate it you should expect suffering.

### DIFF
--- a/src/extra_rules.jl
+++ b/src/extra_rules.jl
@@ -179,8 +179,14 @@ end
 Base.view(t::Tangent{T}, inds) where T<:SVector = view(T(ChainRulesCore.backing(t.data)), inds)
 Base.getindex(t::Tangent{<:SVector, <:NamedTuple}, ind::Int) = ChainRulesCore.backing(t.data)[ind]
 
-function ChainRules.frule((_, ∂x), ::Type{SArray{S, T, N, L}}, x::NTuple{L,Any}) where {S, T, N, L}
-    SArray{S, T, N, L}(x), SArray{S}(∂x)
+function ChainRules.frule(
+    (_, ∂x)::Tuple{Any, Tangent{TUP}},
+    ::Type{SArray{S, T, N, L}},
+    x::TUP,
+) where {L, TUP<:NTuple{L, Number}, S, T<:Number, N}
+    y = SArray{S, T, N, L}(x)
+    ∂y = SArray{S, T, N, L}(ChainRulesCore.backing(∂x))
+    return y, ∂y
 end
 
 @ChainRulesCore.non_differentiable StaticArrays.promote_tuple_eltype(T)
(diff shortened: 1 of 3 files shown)


src/extra_rules.jl
```diff
@@ -179,8 +179,14 @@ end
 Base.view(t::Tangent{T}, inds) where T<:SVector = view(T(ChainRulesCore.backing(t.data)), inds)
 Base.getindex(t::Tangent{<:SVector, <:NamedTuple}, ind::Int) = ChainRulesCore.backing(t.data)[ind]
 
-function ChainRules.frule((_, ∂x), ::Type{SArray{S, T, N, L}}, x::NTuple{L,Any}) where {S, T, N, L}
-    SArray{S, T, N, L}(x), SArray{S}(∂x)
+function ChainRules.frule(
+    (_, ∂x)::Tuple{Any, Tangent{TUP}},
```
staticfloat, line 183:
```suggestion
    (_, ∂x)::Tuple{<:Any, Tangent{TUP}},
```
Do you not need the subtype thing here to avoid diagonal dispatch problems?

test/extra_rules.jl
```diff
@@ -0,0 +1,35 @@
...
+
+    @testset "homogenious type" begin
+        x = (10.0, 20.0, 30.0)
+        ẋ = zero_tangent(x)
+        y, ẏ = frule((NoTangent(), ẋ), StaticArraysCore.SVector{3, Float64}, x)
+        @test y == @SVector [10.0, 20.0, 30.0]
+        @test ẏ == @SVector [0.0, 0.0, 0.0]
+    end
+
+    @testset "convertable type" begin
```
staticfloat, line 18: How is this case different than the first?

src/extra_rules.jl
oxinabox: You do not afaik, because tuple being covarient.
I have run into weird behavour with using `<:` inside tuple type parameters (but in theory it should just have no effect)

test/extra_rules.jl
oxinabox: The first element of `x` is an `Int` rather than an `Float64`

test/extra_rules.jl
```diff
@@ -0,0 +1,35 @@
...
+        x = (10, 20.0, 30.0)
```
oxinabox, line 19:
```suggestion
        x::Tuple{Int, Float64, Float64} = (10, 20.0, 30.0)
```

oxinabox:
cool and I see you have pushed this to the for_cedar branch also

PR #276, opened 2024-03-05
Construct natural tangents for Number and AbstractArray{<:Number} in forward mode

oxinabox:
Ok.
Basically a lot of the code in ChainRules assumes that if your type is a `Number` then its tangent is a number of that same type.
A slightly smaller, but still non-zero portion of code assumed the same about

This is an alternative to 
https://github.com/JuliaDiff/Diffractor.jl/pull/272 and https://github.com/JuliaDiff/ChainRules.jl/pull/787
Since if you *always* work with natural tangent types for `Number`s you don't run into problems like in ability to add them, since that same addition occurs in the primal problem.
And that feels better, mostly.

I think it will also obsolete a lot of hackier things like our special handling for StaticArrays
e.g. what had to be changed in https://github.com/JuliaDiff/Diffractor.jl/pull/275
we shouldn't need that rule at all. And probably some of the other rules in `extra_rules.jl` can go away.

To support Diffractor over ForwardDiff we need *both* handling of natural tangents for `AbstractArray{<Number}` (for `ForwardDiff.Partials`) and handling of natural tangents for `Number` (for `ForwardDiff.Dual`, which uses `ForwarDiff.Partials`)

### DIFF
--- a/src/stage1/recurse_fwd.jl
+++ b/src/stage1/recurse_fwd.jl
@@ -13,6 +13,7 @@ struct ∂☆new{N}; end
 # we split out the 1st order derivative as a special case for performance
 # but the nth order case does also work for this
 function (::∂☆new{1})(B::Type, xs::AbstractTangentBundle{1}...)
+    @info "∂☆new{1}" B supertype(B) typeof(xs)
     primal_args = map(primal, xs)
     the_primal = _construct(B, primal_args)
     tangent_tup = map(first_partial, xs)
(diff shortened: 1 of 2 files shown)


oxinabox:
This should also be called in
`zero_bundle` since `zero_tangent` doesn't do it for AbstractArrays only Arrays

src/stage1/recurse_fwd.jl
```diff
@@ -50,6 +53,28 @@ _construct(::Type{B}, args) where B<:Tuple = B(args)
...
+maybe_construct_natural_tangent(::Type, structural_tangent) = structural_tangent
+for BaseSpaceType in (Number, AbstractArray{<:Number})
+    @eval function maybe_construct_natural_tangent(::Type{B}, structural_tangent) where B<:$BaseSpaceType
+        try
+            # TODO: should this use `_construct` ?
+            # TODO: is this right?
+            unwrap_tup(x::Tangent{<:Tuple}) = ChainRulesCore.backing(x)
+            unwrap_tup(x) = x
+            field_tangents = map(unwrap_tup, structural_tangent)
+            B(field_tangents...)
```
staticfloat, line 66: @Keno can you advise here?

PR #282, opened 2024-03-15
Just call primal on intrinics if all tangents are zero

oxinabox:
Encountered this in the wild.
I can't however reproduce this in a simple test like
```julia
@testset "zeroed tangents intrinstics" begin
    foo(x,y) = Core.Intrinsics.and_int(x, y)
    z = ∂☆internal{1}()(DNEBundle{1}(foo), DNEBundle{1}(true), DNEBundle{1}(false))
    @test primal(z) == foo(true, false)
    @test first_partial
end
```
Not sure why. Maybe the more general handling of all DNEBundle applies there but for some reason in my real world example? 

An alternative to this is to change our general fallback handling of intrinsics to handle zeros, whether compiletime known or not
```julia
function (this::∂☆{N})(f::AbstractZeroBundle{N, Core.IntrinsicFunction}, args::ATB{N}...) where {N}
    ff = primal(f)
    primal_args = map(primal, args)
    primal_out = ff(primal_args...)
    if all(iszero, primal_args))
        (zero_bundle{N}())(primal_out)
    elseif ff in (Base.not_int, Base.ne_float)
        DNEBundle{N}(primal_out)
    else
        error("Missing rule for intrinsic function $ff")
    end
end
```

### DIFF
--- a/src/stage1/forward.jl
+++ b/src/stage1/forward.jl
@@ -285,6 +285,12 @@ function (this::∂☆{N})(::AbstractZeroBundle{N, typeof(typeof)}, x::ATB{N}) w
     DNEBundle{N}(typeof(primal(x)))
 end
 
+
+function (this::∂☆{N})(f::AbstractZeroBundle{N, Core.IntrinsicFunction}, args::AbstractZeroBundle{N}...) where {N}
+    ff=primal(f)
+    return (zero_bundle{N}())(ff(map(primal, args)...))
+end
+
 function (this::∂☆{N})(f::AbstractZeroBundle{N, Core.IntrinsicFunction}, args::ATB{N}...) where {N}
     ff = primal(f)
     if ff in (Base.not_int, Base.ne_float)


Keno:
I've generally been resistant to making this part of the base Diffractor semantics, because people have occasionally written rules that violate this, e.g. the Flux rules for testing whether something is happening under differentiation or the batchnorm update rules. I also don't think it's fully necessary, because stage 2 diffractor should be able to analyze the code and determine whether this is true. That said, if this fixes a short term issue, I'm fine with it for now.

oxinabox:
We have this for things that are not intrinsics and have had for a long time.
Practically it is very useful for avoiding ADing code not on the main path.
This particular case for incidence actually seems less objectionable on that reason.

But yes, moving to traving what is on the active path directly is my preference, longer term.
Not so much for things like batch norm update rules though, I kinda think those shouldn't be done via writing `frule`s
and rather should be using a `ScopedValue`, it feels better to me.

PR #284, opened 2024-03-18
adjustments to the latest master

aviatesk:
Many adjustments...


test/stage2_fwd.jl
```diff
@@ -6,36 +6,34 @@ module stage2_fwd
         @test sin′(1.0) == cos(1.0)
     end
     let sin′′ = Diffractor.dontuse_nth_order_forward_stage2(Tuple{typeof(mysin), Float64}, 2)
-        # This broke some time between 1.10 and 1.11-DEV.10001
-        @test isa(sin′′, Core.OpaqueClosure{Tuple{Float64}, Float64}) broken=VERSION>=v"1.11-"
```
oxinabox, line 10: oh yay!

aviatesk: I don't know too much why these broken test cases have been fixed, but I guess it was probably because the previous `(::∂☆recurse)(...)` implementation has the serious flaw (i.e. the generator uses function that hasn't been defined at the time when the generator function is defined)?

src/codegen/reverse.jl
```diff
@@ -1,16 +1,20 @@
 # Codegen shared by both stage1 and stage2
 
-function make_opaque_closure(interp, typ, name, meth_nargs::Int, isva, lno, cis, revs...)
+function make_opaque_closure(interp, typ, name, meth_nargs::Int, isva, lno, ci, revs...)
     if interp !== nothing
-        cis.inferred = true
+        @static if VERSION ≥ v"1.12.0-DEV.15"
+            rettype = Any # ci.rettype # TODO revisit
```
oxinabox, line 6: Why? is inference broken?

oxinabox (approved):
Pending getting CI working.
This looks good.

src/codegen/reverse.jl
aviatesk: `ci` no longer has `rettype`.

aviatesk: We probably need further tweak on this code, but maybe later.

src/codegen/reverse.jl
```diff
@@ -1,16 +1,20 @@
...
+        else
+            ci.inferred = true
+            rettype = ci.rettype
+        end
         ocm = ccall(:jl_new_opaque_closure_from_code_info, Any, (Any, Any, Any, Any, Any, Cint, Any, Cint, Cint, Any),
```
Keno, line 11: The signature of this c function changed

Keno: That said, this should ideally use the Core.OpaqueClosure constructor

Keno:
This branch now passes tests for me on master with:
- https://github.com/JuliaLang/julia/pull/53875
- https://github.com/JuliaLang/julia/pull/53876
- https://github.com/JuliaLang/julia/pull/53877
- https://github.com/JuliaLang/julia/pull/53878

aviatesk:
I will give them reviews quickly.

aviatesk:
It's really nice to have tests passing on the stable version!

test/stage2_fwd.jl
aviatesk, line 10: These test cases about inferred OC types appear to work fine in v1.12, but they're still broken on v1.11. We might need to consider backporting some changes, or tweak the `broken` condition here.

PR #285, opened 2024-03-23
fix CI setup

aviatesk:


aviatesk:
I've managed to get the CI setup done. As you can see, all tests are currently failing (even including some syntax errors), but this PR aims to highlight the problems that #284 is set to fix.

PR #295, opened 2024-10-02
Adjust forward stage2 to Core.Compiler changes

Keno:
Only what is necessary for Cedar right now. Ordinary stage 2 reverse mode will need similar changes at a later point.

### DIFF
--- a/src/analysis/forward.jl
+++ b/src/analysis/forward.jl
@@ -1,11 +1,71 @@
 using Core.Compiler: StmtInfo, ArgInfo, CallMeta, AbsIntState
 
+if VERSION >= v"1.12.0-DEV.1268"
+
+using Core.Compiler: Future
+
+function fwd_abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
+        arginfo::ArgInfo, si::StmtInfo, sv::AbsIntState, primal_call::Future{CallMeta})
+    if f === ChainRulesCore.frule
+        # TODO: Currently, we don't have any termination analysis for the non-stratified
+        # forward analysis, so bail out here.
+        return primal_call
+    end
+
+    nargs = length(arginfo.argtypes)-1
+    frule_preargtypes = Any[Const(ChainRulesCore.frule), Tuple{Nothing,Vararg{Any,nargs}}]
+    frule_argtypes = append!(frule_preargtypes, arginfo.argtypes)
+    local frule_atype::Any = CC.argtypes_to_type(frule_argtypes)
+
+    local frule_call::Future{CallMeta}
+    local result::Future{CallMeta} = Future{CallMeta}()
+    function make_progress(_, sv)
+        if isa(primal_call[].info, UnionSplitApplyCallInfo)
+            result[] = primal_call[]
+            return true
+        end
+
+        ready = false
+        if !@isdefined(frule_call)
+            # Here we simply check for the frule existance - we don't want to do a full
+            # inference with specialized argtypes and everything since the problem is
+            # likely sparse and we only need to do a full inference on a few calls.
+            # Thus, here we pick `Any` for the tangent types rather than trying to
+            # discover what they are. frules should be written in such a way that
+            # whether or not they return `nothing`, only depends on the non-tangent arguments
+            frule_arginfo = ArgInfo(nothing, frule_argtypes)
+            frule_si = StmtInfo(true)
+            # turn off frule analysis in the frule to avoid cycling
+            interp′ = disable_forward(interp)
+            frule_call = CC.abstract_call_gf_by_type(interp′,
+                ChainRulesCore.frule, frule_arginfo, frule_si, frule_atype, sv, #=max_methods=#-1)::Future
+            isready(frule_call) || return false
+        end
+
+        frc = frule_call[]
+        pc = primal_call[]
+
+        if frc.rt !== Const(nothing)
+            result[] = CallMeta(pc.rt, pc.exct, pc.effects, FRuleCallInfo(pc.info, frc))
+        else
+            result[] = pc
+            CC.add_mt_backedge!(sv, frule_mt, frule_atype)
+        end
+
+        return true
+    end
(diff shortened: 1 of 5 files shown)


src/analysis/forward.jl
```diff
@@ -1,5 +1,65 @@
...
+    end
+
+    nargs = length(arginfo.argtypes)-1
+    frule_preargtypes = Any[Const(ChainRulesCore.frule), Tuple{Nothing,Vararg{Any,nargs}}]
+    frule_argtypes = append!(frule_preargtypes, arginfo.argtypes)
+    frule_atype = CC.argtypes_to_type(frule_argtypes)
+
+    local frule_call::Future{CallMeta}
+    local result::Future{Union{CallMeta, Nothing}} = Future{Union{CallMeta, Nothing}}()
+    function make_progress(_, sv)
```
Keno, line 22: @vtjnash, please confirm that this is the intended way to use this.

aviatesk (approved):
LGTM.

src/analysis/forward.jl
```diff
@@ -1,5 +1,65 @@
...
+    if f === ChainRulesCore.frule
+        # TODO: Currently, we don't have any termination analysis for the non-stratified
+        # forward analysis, so bail out here.
+        return Future{Union{CallMeta, Nothing}}(nothing)
+    end
+
+    nargs = length(arginfo.argtypes)-1
+    frule_preargtypes = Any[Const(ChainRulesCore.frule), Tuple{Nothing,Vararg{Any,nargs}}]
+    frule_argtypes = append!(frule_preargtypes, arginfo.argtypes)
+    frule_atype = CC.argtypes_to_type(frule_argtypes)
```
vtjnash, line 18: Be careful not to closure capture any types, as your performance may suffer quite badly, but still just fast enough you won't notice (e.g. the sysimage could build still when I missed one of these cases, it just took several times longer)

src/analysis/forward.jl
vtjnash: You seem to have closure captured interp instead of using the argument? The interp struct is commonly quite large, so that can increase memory usage quite a bit

Keno: The argument is the wrong interp

src/analysis/forward.jl
```diff
@@ -1,5 +1,65 @@
...
+
+    local frule_call::Future{CallMeta}
+    local result::Future{Union{CallMeta, Nothing}} = Future{Union{CallMeta, Nothing}}()
+    function make_progress(_, sv)
+        if isa(primal_call[].info, UnionSplitApplyCallInfo)
+            result[] = nothing
```
vtjnash, line 24: The return type appears to be wrong here. The intended behavior appears to be returning `result[] = primal_call[]` in this case (https://github.com/JuliaDiff/Diffractor.jl/blob/778af00be4cd90d2d14d8c78055e7d060ccd6286/src/stage2/abstractinterpret.jl#L77-L80)

Keno: That's the function that calls this. Potentially it should be refactored to just do that, but I just wanted to only make the refactoring change.

src/analysis/forward.jl
vtjnash: Yes, this way of defining a `make_progress` seems fine. There isn't really one right answer about how to code this, so base itself already uses probably 3 or 4 different patterns, depending on what kept the original code control flow seemed least distorted. I hadn't used the `@isdefined` trick, but it is essentially equivalent to the nextstate pattern I'd used for manual stackless state machine conversion

src/analysis/forward.jl
Keno: So put it in a Ref{Any}?

vtjnash: Yeah, or a `Core.Box` equivalently. I'd found several places where we had a MethodMatch object that was needed anyways, so that also happened to work sometimes

src/analysis/forward.jl
vtjnash: The return type is required to be Future{CallMeta} though, or the caller's caller will be unhappy

Keno: The call-site there isn't updated yet. This function is called directly from DAECompiler and I adjusted the call-site there to work with `Future{Union{Nothing, CallMeta}}`

src/analysis/forward.jl
vtjnash: I don't think capturing the interp will give you the behavior you want. I think you might need to mutate `sv` instead of re-using `sv` with different `interp`

Keno: I'm not re-using `sv` with a different `interp`. `sv` here is an `IRInterpretationState`, which doesn't have an `interp` argument, so when the callback later gets scheduled, there's just some random interp in there.

src/analysis/forward.jl
vtjnash: It is fine if it still branches on r!==nothing there, it will just be dead code now, as it appears you you must handle that case here now, instead of being able to handle it there

src/analysis/forward.jl
vtjnash: sv claims to be a `AbsIntState` here? For `IRInterpretationState` currently it passes the interp here that was originally used to construct the `IRInterpretationState`, since everything is on the stack there and doesn't handle recursion

I think the behavior here is also probably fine, but that no other callback will be using the right interp, since none other are expecting the interp to be different from the one used to allocate the state object

Keno: > and doesn't handle recursion

Doesn't handle recursion in `Base` ;).

vtjnash: Fair, I know almost nothing about this code, so I am reviewing without really knowing how this integrates. The current implementation in Base would potentially break here: https://github.com/JuliaLang/julia/blob/be401635fe02b28ce994e2e3cae0733d101f8927/base/compiler/ssair/irinterp.jl#L154
since it was not tracking if the return type changed to reschedule this instruction if it became part of cycle (I believe it should detect and `@assert` though if that attempts to happen)

PR #296, opened 2024-11-05
Adjust to upstream edges change

Keno:


Keno:
cc @vtjnash

vtjnash (approved):

src/stage2/lattice.jl
```diff
@@ -66,14 +66,22 @@ end
...
 CC.getsplit_impl(info::FRuleCallInfo, idx::Int) = CC.getsplit(info.info, idx)
 CC.getresult_impl(info::FRuleCallInfo, idx::Int) = CC.getresult(info.info, idx)
 if isdefined(CC, :add_uncovered_edges_impl)
     CC.add_uncovered_edges_impl(edges::Vector{Any}, info::FRuleCallInfo, @nospecialize(atype)) = CC.add_uncovered_edges!(edges, info.info, atype)
 end
+if isdefined(CC, :add_edges_impl)
+    function CC.add_edges_impl(edges::Vector{Any}, info::FRuleCallInfo)
+        CC.add_edges!(edges, info.info)
+        CC.add_edges!(edges, info.frule_call.info)
+    end
```
aviatesk, line 83: I don’t think this implementation allows the method table backedge equivalent to `CC.add_mt_backedge!(sv, frule_mt, frule_atype)` to be created, but I also think this should be sufficient for the purpose of invalidating when a new frule is added—which was the reason for adding `add_mt_backedge!`. In the current implementation, a backedge is added to the fallback `frule(...)`, which I think should trigger equivalent invalidation.

aviatesk (approved):

src/stage2/lattice.jl
Keno: Right, the point was to trigger on additions of new signatures that would change the lookup result. My understanding is that the callinfo encodes exactly the lookup we want, so this should do what we need.

PR #301, opened 2025-03-18
Fix for nightly

serenity4:
Adjusts to a signature change in https://github.com/JuliaLang/julia/pull/57074 as in https://github.com/JuliaDebug/Cthulhu.jl/pull/627, with some cleanup on version checks to reduce complexity (essentially removing support for intermediate development versions, such as between two `-DEV.X` and `-DEV.Y` for 1.11/1.12).


serenity4:
@Keno

